PR #2, opened 2014-06-11
Use type checking to detect an errant closure

evanphx:
Inspired by http://twitter.com/jessemcnelis/status/476704848486731776 I decided to figure out a way to check for the socket closure without comparing strings.

It's a bit longer, but it's more correct.

Side note: I'm thinking maybe about moving this function to a separate package because this is such a common issue.

### DIFF
--- a/session.go
+++ b/session.go
@@ -8,9 +8,9 @@ import (
 	"log"
 	"math"
 	"net"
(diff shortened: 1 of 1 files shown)


evanphx:
bitmoji

PR #8, opened 2015-07-06
session: return explicit nil to avoid non-nil interface value

fatih:
A interface has two elements, a type and a value. A interface value is
only nil if both elements are nil. However if we return `AccepStream()`
directly, net.Conn will be: `(net.Conn, nil)`. Even if the inner value is
nil, this value will be non-nil. So when `Open()` and `Accept()` return an
error, the value (in our example net.Conn) will be always non-nil.

This commit fixes it by explicit returning a nil. More detail about
this can be found: http://golang.org/doc/faq#nil_error

### DIFF
--- a/session.go
+++ b/session.go
@@ -129,7 +129,11 @@ func (s *Session) NumStreams() int {
 
 // Open is used to create a new stream as a net.Conn
 func (s *Session) Open() (net.Conn, error) {
-	return s.OpenStream()
+	conn, err := s.OpenStream()
+	if err != nil {
+		return nil, err
+	}
+	return conn, nil
 }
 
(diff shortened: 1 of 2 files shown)


fatih:
We hit it because we were using `stream.Close()` in a defer to close if something would go wrong in the tunnel package: https://github.com/koding/tunnel

Even If I would check like this:

```go
if stream != nil {
	stream.Close()
}
```

It would panic, because `stream`'s inner value was nil.

mitchellh:
Ah yes, love that "feature" of Go. 

LGTM but will defer to @armon

armon:
Thanks!

PR #9, opened 2015-08-12
Fixes read/write loop and ping deadlocks

slackpad:
If you take out the timeout, the unit test will deadlock when the `Read()` call tries to send the window update. I'm not sure I like the special-case body check in here but I thought I'd push it up so you can take a look (if there's a body we already have an optional deadline interface outside of here).

### DIFF
--- a/session.go
+++ b/session.go
@@ -327,13 +327,19 @@ func (s *Session) waitForSendErr(hdr header, body io.Reader, errCh chan error) e
 	}
 }
 
-// sendNoWait does a send without waiting
+// sendNoWait does a send without waiting. Since there's still a case where
+// sendCh itself can be full, we will enforce the configured HeaderWriteTimeout,
+// since this is a small control header.
 func (s *Session) sendNoWait(hdr header) error {
+	timeout := time.After(s.config.HeaderWriteTimeout)
+
 	select {
 	case s.sendCh <- sendReady{Hdr: hdr}:
 		return nil
 	case <-s.shutdownCh:
 		return ErrSessionShutdown
+	case <-timeout:
+		return ErrHeaderWriteTimeout
 	}
 }
 
@@ -481,8 +487,7 @@ func (s *Session) handlePing(hdr header) error {
 	if flags&flagSYN == flagSYN {
 		hdr := header(make([]byte, headerSize))
 		hdr.encode(typePing, flagACK, 0, pingID)
-		s.sendNoWait(hdr)
-		return nil
+		return s.sendNoWait(hdr)
 	}
 
 	// Handle a response
--- a/session.go
+++ b/session.go
@@ -305,7 +305,10 @@ func (s *Session) waitForSend(hdr header, body io.Reader) error {
 func (s *Session) waitForSendErr(hdr header, body io.Reader, errCh chan error) error {
 	var timeout <- chan time.Time
 	if body == nil {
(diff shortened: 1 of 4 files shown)


slackpad:
An alternative might be to have the window updates managed by a separate goroutine, which would also prevent the deadlock that we are trying to fix here. That seems like overkill, though, because it would be another goroutine and would have to send periodically, possibly causing weirdness and over-flows.

mitchellh:
I'll let @armon have the final say, but this looks like a reasonable solution. The separate goroutine approach scares me from a more "mo parallelism, mo problems" mindset. Given that this seems fairly exceptional as-is, putting it in the critical path shouldn't affect things too much. That is my knee jerk reaction, but I'll think about it more.

session_test.go
```diff
@@ -799,3 +805,58 @@ func TestBacklogExceeded_Accept(t *testing.T) {
 		}
 	}
 }
+
+func TestWindowUpdateWriteDuringRead(t *testing.T) {
+	client, server := testClientServer()
+	defer client.Close()
+	defer server.Close()
+
+	wg := &sync.WaitGroup{}
```
mitchellh: An unimportant nitpick of style: if you're not passing a waitgroup around as a paramter, you can just use `var wg sync.WaitGroup`. Go automatically will turn the receiver into a pointer for function calls, and I think its prettier. But like I said, no functional difference.

session.go
```diff
@@ -299,29 +299,47 @@ func (s *Session) waitForSend(hdr header, body io.Reader) error {
...
+		return ErrHeaderWriteTimeout
 	}
 }
 
-// sendNoWait does a send without waiting
+// sendNoWait does a send without waiting. Since there's still a case where
+// sendCh itself can be full, we will enforce the configured HeaderWriteTimeout,
+// since this is a small control header.
 func (s *Session) sendNoWait(hdr header) error {
+	timeout := time.After(s.config.HeaderWriteTimeout)
```
armon: We should use `time.Timer` and then do a `defer timer.Stop` so that we don't have tons of pending timers

session.go
```diff
@@ -299,29 +299,47 @@ func (s *Session) waitForSend(hdr header, body io.Reader) error {
...
 }
 
-// waitForSendErr waits to send a header, checking for a potential shutdown
+// waitForSendErr waits to send a header with optional data, checking for a
+// potential shutdown. If the body is not supplied then we will enforce the
+// configured HeaderWriteTimeout, since this is a small control header.
 func (s *Session) waitForSendErr(hdr header, body io.Reader, errCh chan error) error {
+	var timeout <- chan time.Time
+	if body == nil {
+		timeout = time.After(s.config.HeaderWriteTimeout)
```
armon: Similarly, we should use a timer and defer the cleanup

session.go
```diff
@@ -465,12 +491,16 @@ func (s *Session) handlePing(hdr header) error {
...
-		s.sendNoWait(hdr)
-		return nil
+		go func() {
+			hdr := header(make([]byte, headerSize))
+			hdr.encode(typePing, flagACK, 0, pingID)
+			if err := s.sendNoWait(hdr); err != nil {
+				s.logger.Printf("[WARN] yamux: failed to send ping reply: %v", err)
+			}
+		}()
 	}
```
armon: We need to keep the `return nil` to avoid processing after starting the goroutine

armon:
LGTM

PR #11, opened 2015-08-19
Generalizes the header write timeout to a broader "safety valve" write timeout.

slackpad:
This opens up the timeout to the user's payloads as well as just control headers. It's a little indirect because we don't have a timeout in the `send()` goroutine directly, but this timeout effectively covers that because the caller is waiting for that goroutine to put the write error onto the error feedback channel.

### DIFF
--- a/const.go
+++ b/const.go
@@ -29,10 +29,6 @@ var (
 	// ErrReceiveWindowExceeded indicates the window was exceeded
 	ErrRecvWindowExceeded = fmt.Errorf("recv window exceeded")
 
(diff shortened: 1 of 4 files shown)


armon:
@slackpad Should we instead try to cast the connection to something with a `SetWriteDeadline` and use deadlines instead? That way we can save a ton of timer setup/teardown throughout.

slackpad:
@armon We could do that by moving the timeouts into session.go's `send()` loop, but then we don't cover the cases where the `sendCh` itself is filled up. I think we want timeout protection on those as well.

armon:
Ah makes sense! :+1:

PR #13, opened 2015-08-23
KeepAliveTimeout: Shuts down the session after timeout

### LINKED ISSUE #12: KeepAlive is not working
Even if Config -> EnableKeepAlive and Config -> KeepAliveInterval are set and the connectivity is lost (i.e. internet goes down), `Session.Accept()` returns no error.
   I'm wondering if it's by design or if it is a bug. If Session.Accept doesn't return any error you can't implement fallback mechanism (e.g. redial the connection), isn't it?

ginkoob:
Closes #12

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -20,8 +20,8 @@ type Config struct {
 	// KeepAliveInterval is how often to perform the keep alive
 	KeepAliveInterval time.Duration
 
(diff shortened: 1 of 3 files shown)


ginkoob:
I'm closing this as it was mixed with the other PR. If there is interest I can  rebase, create a new branch and send a clean patch.

PR #18, opened 2015-09-02
session.go: AcceptStream  handles GoAway

### LINKED ISSUE #17: The server should handle GoAway frames
Currently GoAway is handled only on Session.Open. However there is also the case when the client sends a GoAway request. In that case the server should handle it on Accept(). A common use case is when the the client acts as a "server" (also known as back connection/ reverse connection)

ginkoob:
Closes #17

### DIFF
--- a/session.go
+++ b/session.go
@@ -95,7 +95,6 @@ func newSession(config *Config, conn io.ReadWriteCloser, client bool) *Session {
 		streams:    make(map[uint32]*Stream),
 		synCh:      make(chan struct{}, config.AcceptBacklog),
 		acceptCh:   make(chan *Stream, config.AcceptBacklog),
-		goAwayCh: make(chan struct{}, 1),
 		sendCh:     make(chan sendReady, 64),
 		recvDoneCh: make(chan struct{}),
 		shutdownCh: make(chan struct{}),
@@ -103,6 +102,7 @@ func newSession(config *Config, conn io.ReadWriteCloser, client bool) *Session {
 	if client {
 		s.nextStreamID = 1
 	} else {
+		s.goAwayCh = make(chan struct{})
 		s.nextStreamID = 2
 	}
 	go s.recv()
@@ -203,7 +203,7 @@ func (s *Session) AcceptStream() (*Stream, error) {
 		return stream, nil
 	case <-s.shutdownCh:
 		return nil, s.shutdownErr
-	case <- s.goAwayCh:
+	case <-s.goAwayCh:
 		return nil, ErrRemoteGoAway
 	}
 }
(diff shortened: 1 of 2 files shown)


armon:
@ginkoob So I now remember why it works this way. The `GoAway` is meant to do a half-close. e.g. it indicates to the other side that no new streams will be accepted. Similar to Close in TCP which goes to half-duplex.

ginkoob:
Ok but why does it work only from server to client? Why the client can't say "GoAway" to the server? I think this PR does just that.

session.go
```diff
@@ -92,6 +95,7 @@ func newSession(config *Config, conn io.ReadWriteCloser, client bool) *Session {
 		streams:    make(map[uint32]*Stream),
 		synCh:      make(chan struct{}, config.AcceptBacklog),
 		acceptCh:   make(chan *Stream, config.AcceptBacklog),
+		goAwayCh: make(chan struct{}, 1),
```
slackpad: Little bit of a formatting fail (needs to be indented). Also, `make(chan struct{})` should be sufficient.

session.go
```diff
@@ -416,6 +425,10 @@ func (s *Session) recvLoop() error {
 			handler = s.handleStreamMessage
 		case typeGoAway:
 			handler = s.handleGoAway
+			select{
```
slackpad: I'd move this in to `handleGoAway`.

session_test.go
```diff
@@ -347,6 +347,49 @@ func TestGoAway(t *testing.T) {
...
+		t.Errorf("err: %v", err)
+	}
+	// Test GoAway while Accept is running.
+	client2, server2 := testClientServer()
+	defer client2.Close()
+	defer server2.Close()
+	done = make(chan struct{}, 1)
+	go func(){
+		<- done
+		time.Sleep(500 *time.Millisecond)
```
slackpad: This spacing is strange - I'd probably just run `go fmt` on this.

slackpad:
Noted a couple small things. I see the distinction here - the go away won't cause the server blocked in `Accept` or `AcceptStream` to bail out.

session.go
ginkoob: I think we should create the goAwayCh only if the peer is a server so I've moved it to the if client {}else{}. Let me know if you think otherwise. 
   I've noticed that now session_test.go:375 - `client2.GoAway()` always throws `err: session shutdown`. It shouldn't be an issue as the client shouldn't try to open connections after GoAway but still I don't know why it errors out.

session.go
ginkoob: done!

session_test.go
ginkoob: Done!

slackpad:
Latest code looks good. Only last thing is the fail you mentioned in https://github.com/hashicorp/yamux/pull/18#discussion_r38813184 - I can't reproduce that running your branch locally, and it doesn't make sense given how you've got the delay in there. Are you still seeing that?

evanphx:
Given the age, I'm going to go ahead and close this PR. Per @armon's mention, GoAway wasn't meant to be sent from a client to a server anyway.

evanphx:
@ginkoob If you can sign the CLA, I think we can merge this.

PR #19, opened 2015-09-08
Fixes #12 KeepAlive is not working

### LINKED ISSUE #12: KeepAlive is not working

rnapier:
This adds a timeout of `ConnectionWriteTimeout` for pings. If a keepalive fails, the connection is closed.

Modify test code to turn off keepalives on tests that want to fail for other reasons (since keepalive will tend to fail them faster than the situation under test).

### DIFF
--- a/const.go
+++ b/const.go
@@ -48,6 +48,9 @@ var (
 	// ErrConnectionWriteTimeout indicates that we hit the "safety valve"
 	// timeout writing to the underlying stream connection.
 	ErrConnectionWriteTimeout = fmt.Errorf("connection write timeout")
+
+	// ErrKeepAliveTimeout is sent if a missed keepalive caused the stream close
+	ErrKeepAliveTimeout = fmt.Errorf("keepalive timeout")
 )
 
 const (
--- a/session.go
+++ b/session.go
@@ -272,6 +272,8 @@ func (s *Session) Ping() (time.Duration, error) {
 	start := time.Now()
 	select {
 	case <-ch:
+	case <-time.After(s.config.ConnectionWriteTimeout):
+		return 0, ErrTimeout
 	case <-s.shutdownCh:
 		return 0, ErrSessionShutdown
 	}
@@ -286,7 +288,11 @@ func (s *Session) keepalive() {
 	for {
 		select {
 		case <-time.After(s.config.KeepAliveInterval):
-			s.Ping()
+			_, err := s.Ping()
+			if err != nil {
+				s.logger.Printf("[ERR] yamux: keepalive failed: %v", err)
+				s.Close()
+			}
 		case <-s.shutdownCh:
 			return
 		}
--- a/session.go
+++ b/session.go
@@ -273,6 +273,9 @@ func (s *Session) Ping() (time.Duration, error) {
 	select {
 	case <-ch:
 	case <-time.After(s.config.ConnectionWriteTimeout):
+		s.pingLock.Lock()
+		delete(s.pings, id) // Ignore it if a response comes later.
+		s.pingLock.Unlock()
 		return 0, ErrTimeout
 	case <-s.shutdownCh:
 		return 0, ErrSessionShutdown
@@ -291,7 +294,8 @@ func (s *Session) keepalive() {
 			_, err := s.Ping()
 			if err != nil {
 				s.logger.Printf("[ERR] yamux: keepalive failed: %v", err)
-				s.Close()
+				s.exitErr(ErrKeepAliveTimeout)
+				return
 			}
 		case <-s.shutdownCh:
 			return
--- a/session_test.go
+++ b/session_test.go
@@ -40,12 +40,22 @@ func testConn() (io.ReadWriteCloser, io.ReadWriteCloser) {
 	return conn1, conn2
 }
 
-func testClientServer() (*Session, *Session) {
+func testConf() *Config {
 	conf := DefaultConfig()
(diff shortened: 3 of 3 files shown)


session_test.go
```diff
@@ -981,13 +991,7 @@ func TestSession_PingOfDeath(t *testing.T) {
 }
 
 func TestSession_ConnectionWriteTimeout(t *testing.T) {
-	// Disable keepalives so they don't detect the failed connection
```
slackpad: Good call in making sure these fail for the right reasons. Would you mind adding a test case that exercises your new timeout?

session.go
```diff
@@ -286,7 +288,11 @@ func (s *Session) keepalive() {
 	for {
 		select {
 		case <-time.After(s.config.KeepAliveInterval):
-			s.Ping()
+			_, err := s.Ping()
+			if err != nil {
+				s.logger.Printf("[ERR] yamux: keepalive failed: %v", err)
+				s.Close()
```
slackpad: It would be good to know if `Close()` returns an error and log it to the console.

slackpad:
Hi @rnapier thanks for the PR - I noted a couple small things but this is looking good!

session_test.go
rnapier: Heh; I completely forgot the whole reason I was reworking the tests was so I could add my new test case. Sorry about that… Shortly, yes.

const.go
```diff
@@ -48,6 +48,9 @@ var (
 	// ErrConnectionWriteTimeout indicates that we hit the "safety valve"
 	// timeout writing to the underlying stream connection.
 	ErrConnectionWriteTimeout = fmt.Errorf("connection write timeout")
+
+	// ErrKeepAliveTimeout is sent if a missed keepalive caused the stream close
+	ErrKeepAliveTimeout = fmt.Errorf("keepalive timeout")
```
rnapier: I added a new, explicit error rather than reusing ErrTimeout. I realized that the caller needs to differentiate.

session.go
```diff
@@ -286,7 +291,12 @@ func (s *Session) keepalive() {
 	for {
 		select {
 		case <-time.After(s.config.KeepAliveInterval):
-			s.Ping()
+			_, err := s.Ping()
+			if err != nil {
+				s.logger.Printf("[ERR] yamux: keepalive failed: %v", err)
+				s.exitErr(ErrKeepAliveTimeout)
+				return
+			}
```
rnapier: You mentioned checking the `Close()` error, but I'm not sure what I would do with it. I switched to using your usual pattern of `exitErr()`.

session_test.go
```diff
@@ -5,12 +5,31 @@ import (
...
+func (l *logCapture) match(expect []string) bool {
+	return reflect.DeepEqual(l.logs(), expect)
+}
+
+func captureLogs(s *Session) *logCapture {
+	buf := new(logCapture)
+	s.logger = log.New(buf, "", 0)
+	return buf
+}
+
```
rnapier: I consider logging an important part of the API. I also don't like "expected" errors in the test output. I can add this feature to the other things that currently spit logs if you like.

session.go
slackpad: I was thinking of just logging, but `exitErr` is the way to go.

session_test.go
slackpad: Definitely agree, and thanks! It would be out of scope on this change, but I'd welcome a follow on PR for it.

session_test.go
```diff
@@ -663,6 +734,57 @@ func TestKeepAlive(t *testing.T) {
...
+	select {
+	case err := <-errCh:
+		if err != ErrKeepAliveTimeout {
+			t.Fatalf("unexpected error: %v", err)
+		}
+	case <-time.After(1 * time.Second):
+		t.Fatalf("timeout waiting for timeout")
+	}
+
+	if !client.IsClosed() {
```
slackpad: Running locally it looks like the test is flaky some times right here.

```
workpad:yamux james$ untilfail go test ./...
ok      _/Users/james/projects/rnapier/yamux    12.165s
ok      _/Users/james/projects/rnapier/yamux    11.947s
2015/09/10 10:47:53 [ERR] yamux: keepalive failed: session shutdown
2015/09/10 10:47:53 [ERR] yamux: keepalive failed: session shutdown
--- FAIL: TestKeepAlive_Timeout-4 (0.35s)
        session_test.go:772: client should have closed
2015/09/10 10:48:03 [WARN] yamux: failed to send ping reply: connection write timeout
FAIL
FAIL    _/Users/james/projects/rnapier/yamux    11.800s
```

You could probably remove the client-side checks since all the action is on the server side for this one, or add a small delay for the client to finish the close. I'm puzzled who is emitting `[WARN] yamux: failed to send ping reply: connection write timeout` though, that must be the server but it should be able to do the write.

slackpad:
Changes all LGTM - just one question on there about a little flakyness in one of the unit tests.

rnapier:
Sounds good. I'll fix it up on Monday when I'm back from the mountains.

rnapier:
I think this is now ready to go.

slackpad:
Thanks for this!

PR #21, opened 2015-11-28
Support shrinking of receive buffers of a stream

armon:
This PR adds the `Shrink` method that allows the receive buffers of a Stream to be truncated to zero. This can be useful when connection pooling may keep idle Streams around, as they may have grown their internal buffers to be fairly large (256KB max by default).

/cc: @slackpad

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -439,3 +439,14 @@ func (s *Stream) SetWriteDeadline(t time.Time) error {
 	s.writeDeadline = t
 	return nil
 }
+
+// Shrink is used to compact the amount of buffers utilized
+// This is useful when using Yamux in a connection pool to reduce
+// the idle memory utilization.
+func (s *Stream) Shrink() {
+	s.recvLock.Lock()
+	if s.recvBuf != nil && s.recvBuf.Len() == 0 {
+		s.recvBuf = nil
+	}
+	s.recvLock.Unlock()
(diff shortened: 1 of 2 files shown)


stream.go
```diff
@@ -397,7 +397,12 @@ func (s *Stream) readData(hdr header, flags uint16, conn io.Reader) error {
 
 	// Copy into buffer
 	s.recvLock.Lock()
-	if _, err := io.Copy(&s.recvBuf, conn); err != nil {
+	if s.recvBuf == nil {
+		// Allocate the receive buffer just-in-time to fit the full data frame.
+		// This way we can read in the whole packet without further allocations.
+		s.recvBuf = bytes.NewBuffer(make([]byte, 0, length))
```
slackpad: Interesting that Go has a hard-coded 32k here the way we were doing it before.

slackpad:
LGTM

jbenet:
@armon wonder if pool size + time based auto-shrink would be a good idea 

```go
if len(streams) > k && (now - s.lastRecvd) > timeout { 
  s.Shrink() 
}
```

armon:
@jbenet it's hard to do automatically at the yamux layer as we have no context from the application. We've used this in the Consul and Nomad connection pools however where we do have more context!

Issue #26, opened 2016-03-08
OpenStream hanging on synCh


whyrusleeping:
over on the ipfs project we updated the version of yamux we were using to the latest master from 9feabe6 about a month ago. Since then we started noticing random hanging issues, which i got around to debugging today and it looks like the culprit is the `synCh` channel on the session. Commits that don't contain that logic work just fine, but every commit i've tried with it has hung.

It looks to be an issue with not replacing tokens in the semaphore in all needed cases, maybe a call to `Close` isnt actually releasing a stream the way it should?

Increasing the `AcceptBacklog` in the config 'fixes' the problem, but that just seems like its just buying me more time until it hangs again.

armon:
@whyrusleeping Can you verify that the accept backlog isn't being reached? e.g. it is expected to hang if the backlog is full, but you are saying that it is hanging when the acceptor has not reached the backlog. Do both sides have the same backlog configured (I think that is a potential issue if the two sides disagree).

whyrusleeping:
@armon i'm pretty sure the accept backlog isnt being reached, after reporting the issue i bumped the number up to 8192, which made the issue less likely to occur, but I still encounter it.

I'm reasonably certain that both sides are using the same values for accept backlog. we recently made a breaking change to our network protocol requiring everyone to update to the latest code in order to be on the network, and I still see the hang happening. Here is a stack trace if you are interested:

https://ipfs.io/ipfs/QmUnwtP8rQaVFKdF2gFb2VstiLmySAXByvHG2KzZ4xGqZ6

whyrusleeping:
Any update here? I hit this issue rather frequently

whyrusleeping:
correction to the situation: We're likely hitting the accept backlog limit, but hitting that shouldnt cause it to permanently hang. It should just act like a semaphore and hang until something else completes

whyrusleeping:
I believe i've found the issue. If you open a stream, write to it, then close it fast enough, you can close it before its 'established' thus causing the synCh semaphore to not be released.

Heres a repro:
```go
package main

import (
	"fmt"
	"net"
	"time"

	"github.com/hashicorp/yamux"
)

func tcppipe() (net.Conn, net.Conn) {
	list, err := net.Listen("tcp", "0.0.0.0:0")
	if err != nil {
		panic(err)
	}

	con, err := net.Dial("tcp", list.Addr().String())
	if err != nil {
		panic(err)
	}

	con2, err := list.Accept()
	if err != nil {
[...]

PR #28, opened 2016-05-11
properly handle closing streams before they are ACKed

whyrusleeping:
Should address bug #26. The issue occurs when opening unidirectional streams then closing them quickly before the other side can ACK the open.

### DIFF
--- a/session_test.go
+++ b/session_test.go
@@ -148,6 +148,47 @@ func TestPing_Timeout(t *testing.T) {
 	}
 }
 
+func TestCloseBeforeAck(t *testing.T) {
+	cfg := testConf()
+	cfg.AcceptBacklog = 8
+	client, server := testClientServerConfig(cfg)
+
+	defer client.Close()
+	defer server.Close()
+
+	for i := 0; i < 8; i++ {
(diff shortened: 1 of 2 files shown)


whyrusleeping:
After further testing I can still reproduce the bug (though it requires far more effort) in ipfs. In the small reproduction program posted in the tracking issue I cannot get a repro.

whyrusleeping:
Alright, got it figured out properly. Ready for review

whyrusleeping:
one potential issue is that this can still lock up if one side doesnt have the fix. (since they won't ACK correctly). This could be addressed either by requiring everyone to use the correct version, or maybe by adding a timeout to that select case in session.go:149

slackpad:
Hi @whyrusleeping thanks for the PR and sorry for the delay - finally had a little time to go through this in some detail. I think there should be a way to do this completely on the sender side without relying on the other party sending an ACK to clear the semaphore. I'll work up a change based on your PR and post it here shortly.

whyrusleeping:
@slackpad thanks for getting that in. Not relying on the other side to be correct is definitely a better solution.

PR #29, opened 2016-05-18
Properly handles closing streams before they are ACKed.

### LINKED ISSUE #26: OpenStream hanging on synCh

slackpad:
This takes #28 which fixes #26 but puts all the logic on the side opening the stream, so we don't depend on any updated logic or messages from the other side. In addition:

* Modified the logic to clean up the `synCh` when a stream is established (like before) and any time a stream is closed. This latter change is easier to reason about, and we don't have to trace through every possible close scenario to make sure we catch them all.
* Changed the `panic()`s in the `synCh` handling code to `ERR` logs, since underflowing that channel _could_ happen from a misbehaving client on the other side sending extra ACKs. Also, this is recoverable, so we don't need to panic.
* Added some cleanup code for `synCh` if the initial window update can't be sent.

### DIFF
--- a/session.go
+++ b/session.go
@@ -178,7 +178,7 @@ GET_ID:
 		select {
 		case <-s.synCh:
 		default:
-			s.logger.Printf("[ERR] aborted stream open without inflight syn semaphore")
+			s.logger.Printf("[ERR] yamux: aborted stream open without inflight syn semaphore")
 		}
 		return nil, err
 	}
@@ -598,7 +598,7 @@ func (s *Session) closeStream(id uint32) {
 		select {
 		case <-s.synCh:
 		default:
-			s.logger.Printf("[ERR] un-established stream without inflight syn semaphore")
+			s.logger.Printf("[ERR] yamux: un-established stream without inflight syn semaphore")
 		}
 	}
 	delete(s.streams, id)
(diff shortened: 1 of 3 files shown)


session.go
```diff
@@ -166,10 +170,16 @@ GET_ID:
...
 	s.streams[id] = stream
+	s.inflight[id] = struct{}{}
 	s.streamLock.Unlock()
 
 	// Send the window update to create
 	if err := stream.sendWindowUpdate(); err != nil {
+		select {
+		case <-s.synCh:
+		default:
+			s.logger.Printf("[ERR] aborted stream open without inflight syn semaphore")
```
ryanuber: Can we add `yamux:` to all of these log messages?

session.go
```diff
@@ -580,19 +590,34 @@ func (s *Session) incomingStream(id uint32) error {
...
-// issued a close.
+// issued a close. If there was an in-flight SYN and the stream
+// was not yet established, then this will give the credit back.
 func (s *Session) closeStream(id uint32) {
 	s.streamLock.Lock()
+	if _, ok := s.inflight[id]; ok {
+		select {
+		case <-s.synCh:
+		default:
+			s.logger.Printf("[ERR] un-established stream without inflight syn semaphore")
```
ryanuber: Would it cover it to say `close of un-established stream` ? Also, is this an `ERR`, or should it be a `WARN` ?

slackpad: @ryanuber I tweaked the wording on these a bit. Since any of these errors are either logic errors or a malicious client I figured I'd log them with `ERR` visibility since it's fairly serious.

PR #31, opened 2016-06-08
use a timer instead of 'After' to avoid leaking resources

whyrusleeping:
Calling time.After creates a runtime timer that gets scheduled into the runtime. Doing this frequently will bog the scheduler down and have weird effects on the program. Using a timer instead will let us stop the timer and free up those resources when we exit the function

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -118,12 +118,17 @@ START:
 
 WAIT:
 	var timeout <-chan time.Time
+	var timer *time.Timer
 	if !s.readDeadline.IsZero() {
 		delay := s.readDeadline.Sub(time.Now())
-		timeout = time.After(delay)
+		timer = time.NewTimer(delay)
+		timeout = timer.C
 	}
 	select {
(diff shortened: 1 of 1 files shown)


whyrusleeping:
cc @slackpad

sean-:
What does a `time.Timer` leak?  `time.Ticker` can leak a resource.  ?

whyrusleeping:
@sean- the `time.Timer` doesnt leak anything so long as you either `Stop` it, or it fires. `time.Ticker` will continue running until you call `Stop` on it. The issue here is that if we have a large number of rapid calls to read and a timeout is set, there are going to be *many* inflight `time.After` timers running since they don't get explicitly cancelled.

armon:
LGTM

lattwood:
The write side of this functionality has been added to hashicorp/yamux#127, as well as a bug fix around ensuring stopped Timers have their channels drained when appropriate.

PR #33, opened 2016-06-14
lock around buffer check on close

whyrusleeping:
Small race condition detected in the race detector. Not sure if it caused any issues or not


armon:
Fixed by #34. Thanks!

PR #34, opened 2016-07-02
Fix race for recvBuf

erikdubbelboer:
This is a race condition found running my own code with `-race`. Running the tests for this library with `-race` results in many, what I think are probably not important, race conditions and fails some tests. So writing a test case for this is impossible with the current state of the tests.

From what I can see no other place tries to lock `recvLock` and `stateLock` at the same time so this shouldn't introduce any deadlocks.

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -91,10 +91,13 @@ START:
 	case streamRemoteClose:
 		fallthrough
 	case streamClosed:
+		s.recvLock.Lock()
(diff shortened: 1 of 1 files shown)


armon:
@slackpad want to give this a look over? LGTM

slackpad:
This LGTM as well, merging.

slackpad:
@erikdubbelboer thanks!

PR #35, opened 2016-07-28
Keep RTT if you are already doing keepalive

hagna:
Suppose we keep track of RTT?


m4ng0squ4sh:
This causes a race, if not protected with a mutex... Make it private and create a getter function which handles the synchronization.

evanphx:

PR #38, opened 2016-08-20
optimize ping : if the data is flowing into recvLoop, ping is not required.

xtaci:
consider the following situation:
1. a link has 200KB/s bandwidth, and the underlying socket buffer is 4MB(large enough)
2. set yamux.Config MaxStreamWindowSize to sockbuf(large enough)
3. constantly send data via streams, eg : 2MB/s to make the socket buffer full
4. then the keepalive cannot respond in 10seconds ( 2MB/200KB)

I think we don't need to send ping packets when the data is flowing to the receiver.

### DIFF
--- a/session.go
+++ b/session.go
@@ -45,6 +45,7 @@ type Session struct {
 	pings    map[uint32]chan struct{}
 	pingID   uint32
 	pingLock sync.Mutex
+	needPing int32
 
 	// streams maps a stream id to a stream, and inflight has an entry
 	// for any outgoing stream that has not yet been established. Both are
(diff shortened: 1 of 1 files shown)


session.go
```diff
@@ -286,7 +287,10 @@ func (s *Session) Ping() (time.Duration, error) {
 		s.pingLock.Lock()
 		delete(s.pings, id) // Ignore it if a response comes later.
 		s.pingLock.Unlock()
-		return 0, ErrTimeout
+		// Double check the needPing flag when timeout
+		if needPing := atomic.LoadInt32(&s.needPing); needPing == 0 {
```
AudriusButkevicius: `if atomic.LoadInt32(&s.needPing) == 0 {` would be cleaner

session.go
```diff
@@ -446,6 +452,9 @@ func (s *Session) recvLoop() error {
 		if err := handler(hdr); err != nil {
 			return err
 		}
+
+		// Ping is not required if data is flowing
+		atomic.StoreInt32(&s.needPing, 1)
```
AudriusButkevicius: I don't see it being set back to 0 ever?
What if I had a lot of data flowing but then the stream goes silent?

PR #39, opened 2016-08-20
optimize ping : if the data is flowing into recvLoop, ping is not required

xtaci:
consider the following situation:
1. a link has 200KB/s bandwidth, and the underlying socket buffer is 4MB(large enough)
2. set yamux.Config MaxStreamWindowSize to sockbuf(large enough)
3. constantly send data via streams, eg : 2MB/s to make the socket buffer full
4. then the keepalive cannot respond in 10seconds ( 2MB/200KB)

I think we don't need to send ping packets when the data is flowing to the receiver.

### DIFF
--- a/session.go
+++ b/session.go
@@ -288,7 +288,7 @@ func (s *Session) Ping() (time.Duration, error) {
 		delete(s.pings, id) // Ignore it if a response comes later.
 		s.pingLock.Unlock()
 		// Double check the needPing flag when timeout
-		if needPing := atomic.LoadInt32(&s.needPing); needPing == 0 {
+		if atomic.LoadInt32(&s.needPing) == 0 {
 			return 0, ErrTimeout
 		}
 	case <-s.shutdownCh:
@@ -305,13 +305,15 @@ func (s *Session) keepalive() {
 	for {
 		select {
 		case <-time.After(s.config.KeepAliveInterval):
-			if needPing := atomic.LoadInt32(&s.needPing); needPing == 0 {
+			if atomic.LoadInt32(&s.needPing) == 0 {
 				_, err := s.Ping()
 				if err != nil {
 					s.logger.Printf("[ERR] yamux: keepalive failed: %v", err)
 					s.exitErr(ErrKeepAliveTimeout)
 					return
 				}
+			} else {
+				atomic.StoreInt32(&s.needPing, 0)
 			}
 		case <-s.shutdownCh:
 			return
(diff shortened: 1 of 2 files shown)


xtaci:
@AudriusButkevicius which line?

AudriusButkevicius:
```
+			if atomic.LoadInt32(&s.needPing) == 0 {
 +				_, err := s.Ping()
 +				if err != nil {
 +					s.logger.Printf("[ERR] yamux: keepalive failed: %v", err)
 +					s.exitErr(ErrKeepAliveTimeout)
 +					return
 +				}
 +			} else {
 +				atomic.StoreInt32(&s.needPing, 0)
  			}
```

becomes

```
+			if !atomic.CompareAndSwapInt32(&s.needPing, 1, 0) {
 +				_, err := s.Ping()
 +				if err != nil {
 +					s.logger.Printf("[ERR] yamux: keepalive failed: %v", err)
 +					s.exitErr(ErrKeepAliveTimeout)
 +					return
 +				}
  			}
```

xtaci:
cool! that's much simpler! any more improvements?

xtaci:
@AudriusButkevicius Looks good!

AudriusButkevicius:
Not really.
I guess the regression here is that Ping() will return 0ms if the data is flowing.

xtaci:
@AudriusButkevicius RTT is ignored for the moment

session.go|309 col 17| _, err := s.Ping()

AudriusButkevicius:
Actually, it's might be fine.

If someone calls s.Ping(), the packet is still sent out (given if bandwidth is available). If you get a response, you get an RTT., but if it never gets delivered back, and we have data flowing and needPing is not set, the RTT returned will be KeepaliveTimeout. Not sure if that's an issue.

xtaci:
@AudriusButkevicius and I think Ping() can never makes a good estimation on RTT.

AudriusButkevicius:
well it's some fake RTT, yet still, it's a change in behaviour.

xtaci:
but if we don't do DOUBLE CHECK at line 291, a misjudgement on timeout will happen when :
a burst transmission for a large block of data is on the way to client, and the ping response is appended to the very behind of the large block of data.

AudriusButkevicius:
yeah true, ok so  it's better than what it is now.

xtaci:
@AudriusButkevicius is manually calling Ping() a good idea?

AudriusButkevicius:
Externally or from the code? From the code I guess yes, as it's already doing that. Externally it's down to the user.

PR #47, opened 2016-12-19
Add timeout support to stream acceptance

pjvds:
Add timeout support to stream acceptance:

``` Go
stream, err := s.AcceptStreamWithTimeout(50 * time.Millisecond)
```

We use this to accept streams within a certain time or close the connection. It will block forever when called with an zero duration. Timers will be closed so it is recovered by the garbage collector even if the timer has not been fired yet.


evanphx:
I'd prefer to handle this with #80 since a context can be enhanced with a timeout itself.

PR #48, opened 2017-01-22
Add a callback for keepalive failure

wjh:
I've found this useful to notify my code of a disconnect, and it actually means I don't need to implement my own heartbeat/keepalive (which seems pointless considering yamux already does)


evanphx:

Issue #43, opened 2016-10-18
Stream receivebuffer continually grows


Matthias247:
The library currently uses bytes.Buffer for buffering data that was received from the network connection until it was fetched by the user through a `Read()` call.
The problem with that is that all writes from the network will append to the end of the buffer. Reads will read from the current position and shift the read pointer to the end of the buffer, but the already read data will still stay at the beginning of the buffer - until the stream is closed.

This means e.g. for transferring a 100MB file through yamux the receivebuffer for the stream will grow to 100MB.

This could be fixed by utilizing a ring buffer, which is big enough to hold the maximum window size. Or a buffer of window size where each read copies remaining data to the front, resets the read pointer to 0 and resets the writer pointer according to it.

havoc-io:
It's not entirely true that data will stay at the beginning of the buffer until the stream is closed, is it? If you reach a point where the `bytes.Buffer` is completely emptied, its internal position tracker will be reset to point to the beginning of its buffer and its capacity will be reused (https://golang.org/src/bytes/buffer.go?s=8509:8559#L248).

But I agree that it does seem like even if you are constantly `Read`ing (e.g. doing an `io.Copy` call into a file) you might never actually reach that condition unless you happen to do just the right read with just the right buffer size, and thus the buffer would continue to grow.  Is that what you see in practice?  I ask as someone considering using this library.

Matthias247:
@havoc-io Yes, Buffer has this property. However if you look at https://github.com/hashicorp/yamux/blob/master/stream.go#L107-L112 you will see that this branch will never be taken as `Read()` will only be called if there is something in the buffer. But even if it would apply, as we can't guarantee that the reader is faster than the writer and that the buffer will ever be empty it would not be the best solution.

You can see the problem happens in the unit tests if you output the capacity of the buffer after `TestSendData_Large` with `fmt.Printf("Recv buf cap: %d\n", stream.recvBuf.Cap())`

I would fix it by replacing the Buffer with a simple slice with a behavior similar to here: https://github.com/bradfitz/http2/blob/master/buffer.go

havoc-io:
Ah, that's true. It's seems like a weird design decision on the part of `bytes.Buffer` to have truncation triggered on the beginning of `Read` rather than the end. Triggering at the end would at least potentially truncate in this case, though of course something like a ring buffer would make more sense. Thanks for clarifying!

stuartcarnie:
[...]

PR #50, opened 2017-09-12
Fix window update, reduce memory utilization and improve performance

stuartcarnie:
improve memory utilization in receive buffer, fix flow control

* flow control was assuming a `Read` consumed the entire buffer
* introduced fixed receive buffer on streams to reduce memory
  utilization when receiving large streams of data
* use timer pool to reduce allocations
* use static handler function pointer to avoid closure allocations
  for every frame

---

Obligatory benchmarks

```
benchmark                    old ns/op     new ns/op     delta
BenchmarkSendRecv-2          3942          3655          -7.28%
BenchmarkSendRecvLarge-2     28408768      23931960      -15.76%

benchmark                    old allocs     new allocs     delta
BenchmarkSendRecv-2          6              2              -66.67%
BenchmarkSendRecvLarge-2     4012           1236           -69.19%

benchmark                    old bytes     new bytes     delta
BenchmarkSendRecv-2          302           83            -72.52%
BenchmarkSendRecvLarge-2     1538987       57536         -96.26%
```

### DIFF
--- a/session.go
+++ b/session.go
@@ -323,8 +323,13 @@ func (s *Session) waitForSend(hdr header, body io.Reader) error {
 // potential shutdown. Since there's the expectation that sends can happen
 // in a timely manner, we enforce the connection write timeout here.
 func (s *Session) waitForSendErr(hdr header, body io.Reader, errCh chan error) error {
-	timer := time.NewTimer(s.config.ConnectionWriteTimeout)
-	defer timer.Stop()
+	t := timerPool.Get()
+	timer := t.(*time.Timer)
+	timer.Reset(s.config.ConnectionWriteTimeout)
+	defer func() {
+		timer.Stop()
+		timerPool.Put(t)
+	}()
 
 	ready := sendReady{Hdr: hdr, Body: body, Err: errCh}
 	select {
@@ -408,11 +413,19 @@ func (s *Session) recv() {
 	}
 }
 
+var (
+	handlers = []func(*Session, header) error{
+		typeData:         (*Session).handleStreamMessage,
+		typeWindowUpdate: (*Session).handleStreamMessage,
+		typePing:         (*Session).handlePing,
+		typeGoAway:       (*Session).handleGoAway,
+	}
+)
+
 // recvLoop continues to receive data until a fatal error is encountered
 func (s *Session) recvLoop() error {
 	defer close(s.recvDoneCh)
 	hdr := header(make([]byte, headerSize))
-	var handler func(header) error
 	for {
 		// Read the header
 		if _, err := io.ReadFull(s.bufRead, hdr); err != nil {
@@ -428,22 +441,12 @@ func (s *Session) recvLoop() error {
 			return ErrInvalidVersion
 		}
 
-		// Switch on the type
-		switch hdr.MsgType() {
-		case typeData:
-			handler = s.handleStreamMessage
-		case typeWindowUpdate:
-			handler = s.handleStreamMessage
-		case typeGoAway:
-			handler = s.handleGoAway
-		case typePing:
-			handler = s.handlePing
-		default:
+		mt := hdr.MsgType()
+		if mt < typeData || mt > typeGoAway {
 			return ErrInvalidMsgType
 		}
 
-		// Invoke the handler
-		if err := handler(hdr); err != nil {
+		if err := handlers[mt](s, hdr); err != nil {
 			return err
 		}
 	}
--- a/session.go
+++ b/session.go
@@ -442,7 +442,7 @@ func (s *Session) recvLoop() error {
 		}
 
 		mt := hdr.MsgType()
-		if mt < typeData || mt > typeGoAway {
+		if mt < typeData || mt >= typeMax {
 			return ErrInvalidMsgType
 		}
 
--- a/stream.go
+++ b/stream.go
@@ -238,18 +238,25 @@ func (s *Stream) sendWindowUpdate() error {
 
 	// Determine the delta update
 	max := s.session.config.MaxStreamWindowSize
-	delta := max - atomic.LoadUint32(&s.recvWindow)
+	var bufLen uint32
+	s.recvLock.Lock()
+	if s.recvBuf != nil {
+		bufLen = uint32(s.recvBuf.Len())
+	}
+	delta := (max - bufLen) - s.recvWindow
 
 	// Determine the flags if any
 	flags := s.sendFlags()
 
 	// Check if we can omit the update
 	if delta < (max/2) && flags == 0 {
+		s.recvLock.Unlock()
 		return nil
 	}
 
 	// Update our window
-	atomic.AddUint32(&s.recvWindow, delta)
+	s.recvWindow += delta
+	s.recvLock.Unlock()
 
 	// Send the header
 	s.controlHdr.encode(typeWindowUpdate, flags, s.id, delta)
@@ -392,16 +399,18 @@ func (s *Stream) readData(hdr header, flags uint16, conn io.Reader) error {
 	if length == 0 {
 		return nil
 	}
-	if remain := atomic.LoadUint32(&s.recvWindow); length > remain {
-		s.session.logger.Printf("[ERR] yamux: receive window exceeded (stream: %d, remain: %d, recv: %d)", s.id, remain, length)
-		return ErrRecvWindowExceeded
-	}
 
 	// Wrap in a limited reader
 	conn = &io.LimitedReader{R: conn, N: int64(length)}
 
 	// Copy into buffer
 	s.recvLock.Lock()
+
+	if length > s.recvWindow {
+		s.session.logger.Printf("[ERR] yamux: receive window exceeded (stream: %d, remain: %d, recv: %d)", s.id, s.recvWindow, length)
+		return ErrRecvWindowExceeded
+	}
+
 	if s.recvBuf == nil {
 		// Allocate the receive buffer just-in-time to fit the full data frame.
 		// This way we can read in the whole packet without further allocations.
@@ -414,7 +423,7 @@ func (s *Stream) readData(hdr header, flags uint16, conn io.Reader) error {
 	}
 
 	// Decrement the receive window
-	atomic.AddUint32(&s.recvWindow, ^uint32(length-1))
+	s.recvWindow += ^uint32(length - 1)
 	s.recvLock.Unlock()
 
 	// Unblock any readers
--- a/util.go
+++ b/util.go
@@ -1,5 +1,20 @@
 package yamux
 
+import (
+	"sync"
+	"time"
+)
+
+var (
+	timerPool = &sync.Pool{
+		New: func() interface{} {
+			timer := time.NewTimer(time.Hour * 1e6)
+			timer.Stop()
+			return timer
+		},
+	}
+)
+
 // asyncSendErr is used to try an async send of an error
 func asyncSendErr(ch chan error, err error) {
 	if ch == nil {
--- a/bench_test.go
+++ b/bench_test.go
@@ -79,3 +79,46 @@ func BenchmarkSendRecv(b *testing.B) {
 	}
 	<-doneCh
 }
+
+func BenchmarkSendRecvLarge(b *testing.B) {
+	client, server := testClientServer()
+	defer client.Close()
+	defer server.Close()
+
+	const sendSize = 100 * 1024 * 1024
+	const recvSize = 4 * 1024
+
+	sendBuf := make([]byte, sendSize)
+	recvBuf := make([]byte, recvSize)
+
+	b.ResetTimer()
+	recvDone := make(chan struct{})
+
+	go func() {
+		stream, err := server.AcceptStream()
+		if err != nil {
+			return
+		}
+		defer stream.Close()
+		for i := 0; i < b.N; i++ {
+			for j := 0; j < sendSize/recvSize; j++ {
+				if _, err := stream.Read(recvBuf); err != nil {
+					b.Fatalf("err: %v", err)
+				}
+			}
+		}
+		close(recvDone)
+	}()
+
+	stream, err := client.Open()
+	if err != nil {
+		b.Fatalf("err: %v", err)
+	}
+	defer stream.Close()
+	for i := 0; i < b.N; i++ {
+		if _, err := stream.Write(sendBuf); err != nil {
+			b.Fatalf("err: %v", err)
+		}
+	}
+	<-recvDone
+}
--- a/const.go
+++ b/const.go
@@ -76,6 +76,10 @@ const (
 	// GoAway is sent to terminate a session. The StreamID
 	// should be 0 and the length is an error code.
 	typeGoAway
+
+	// typeMax defines the upper bound of valid message types
+	// and should always be the last constant.
+	typeMax
 )
 
 const (
--- a/session_test.go
+++ b/session_test.go
@@ -376,7 +376,12 @@ func TestSendData_Large(t *testing.T) {
 	defer client.Close()
 	defer server.Close()
 
-	data := make([]byte, 512*1024)
+	const (
+		sendSize = 250 * 1024 * 1024
+		recvSize = 4 * 1024
+	)
+
+	data := make([]byte, sendSize)
 	for idx := range data {
 		data[idx] = byte(idx % 256)
 	}
@@ -390,16 +395,17 @@ func TestSendData_Large(t *testing.T) {
 		if err != nil {
 			t.Fatalf("err: %v", err)
 		}
-
-		buf := make([]byte, 4*1024)
-		for i := 0; i < 128; i++ {
+		var sz int
+		buf := make([]byte, recvSize)
+		for i := 0; i < sendSize/recvSize; i++ {
 			n, err := stream.Read(buf)
 			if err != nil {
 				t.Fatalf("err: %v", err)
 			}
-			if n != 4*1024 {
+			if n != recvSize {
 				t.Fatalf("short read: %d", n)
 			}
+			sz += n
 			for idx := range buf {
 				if buf[idx] != byte(idx%256) {
 					t.Fatalf("bad: %v %v %v", i, idx, buf[idx])
@@ -410,6 +416,8 @@ func TestSendData_Large(t *testing.T) {
 		if err := stream.Close(); err != nil {
 			t.Fatalf("err: %v", err)
 		}
+
+		t.Logf("cap=%d, n=%d\n", stream.recvBuf.Cap(), sz)
 	}()
 
 	go func() {
@@ -439,7 +447,7 @@ func TestSendData_Large(t *testing.T) {
 	}()
 	select {
 	case <-doneCh:
-	case <-time.After(time.Second):
+	case <-time.After(5 * time.Second):
 		panic("timeout")
 	}
 }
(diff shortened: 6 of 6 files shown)


stuartcarnie:
This also addresses issue #43 

The `Large` benchmark now caps out at 512k

stuartcarnie:
[This gist][0] shows real-world numbers for Influx Enterprise

**tl;dr**

* total allocations went from 8.02GB down to 2.87MB or a 2,808 times improvement
* in use memory (for large stream) went from 4.01GB down to 815.27KB or a 39,361 times improvement

[0]: https://gist.github.com/stuartcarnie/6ee099d81f03a8f93524c52a8da84fc3

session.go
```diff
@@ -428,22 +441,12 @@ func (s *Session) recvLoop() error {
...
-			handler = s.handlePing
-		default:
+		mt := hdr.MsgType()
+		if mt < typeData || mt > typeGoAway {
 			return ErrInvalidMsgType
 		}
 
-		// Invoke the handler
-		if err := handler(hdr); err != nil {
+		if err := handlers[mt](s, hdr); err != nil {
```
erikdubbelboer: I don't really like this change. Yes it's less lines of code but it's also much more susceptible to subtle bugs where you forget to change the if condition above. With the switch statement adding or removing a case only affects 1 place while after this change you need to update both the if statement and the lookup table.

stuartcarnie: I made this change to avoid allocations for every message that is handled.

I tried this approach first, however it still resulted in allocations:

```go
switch mt {
case typeData:
    handler = (*Session).handleStreamMessage
case typeWindowUpdate:
    handler = (*Session).handleStreamMessage
...
}

if err := handler(s, hdr)
```

stuartcarnie:
I will add a upper message type constant, so that the `if` won’t need changing when validating the range and how this issue is typically addressed.

session.go
stuartcarnie: See commit e7f9152

erikdubbelboer: Are you sure it allocates? I can't seem to produce any similar code that actually allocates: https://gist.github.com/erikdubbelboer/53f4bc902563293ffa9e3a351ff4a149

If it really allocates I would have turned handlers into a map so you can easily add and remove things while keeping the if the same. But of course this would be a bit slower than what you write now. As long as there are enough tests I guess your typeMax solution is also good.

stuartcarnie:
Here is the original code, where you will notice the call to `newobject`, because the compiler is capturing `s` to make `handler`

```asm
    431            .          . 		// Switch on the type 
    432            .          . 		switch hdr.MsgType() { 
    433            .          . 		case typeData: 
    434          3MB        3MB 			handler = s.handleStreamMessage 
                   .          .  1109973: MOVQ AX, 0(SP)	                                  session.go:434
                 3MB        3MB  1109977: CALL runtime.newobject(SB)	                      session.go:434
                   .          .  110997c: MOVQ 0x8(SP), AX	                                session.go:434
                   .          .  1109981: LEAQ 0x11f98(IP), CX	                            session.go:434
                   .          .  1109988: MOVQ CX, 0(AX)	                                  session.go:434
                   .          .  110998b: TESTB AL, 0(AX)	                                 session.go:434
                   .          .  110998d: MOVL 0x14a4dd(IP), DX	                           session.go:434
                   .          .  1109993: LEAQ 0x8(AX), BX	                                session.go:434
                   .          .  1109997: TESTL DX, DX	                                    session.go:434
                   .          .  1109999: JNE 0x1109a05	                                   session.go:434
                   .          .  110999b: MOVQ 0xb8(SP), BX	                               session.go:434
                   .          .  11099a3: MOVQ BX, 0x8(AX)	                                session.go:434
                   .          .  11099a7: MOVQ 0x50(SP), CX	                               session.go:434
                   .          .  1109a0d: MOVQ BX, 0(SP)	                                  session.go:434
                   .          .  1109a11: MOVQ 0xb8(SP), AX	                               session.go:434
                   .          .  1109a19: MOVQ AX, 0x8(SP)	                                session.go:434
                   .          .  1109a1e: CALL runtime.writebarrierptr(SB)	                session.go:434
                   .          .  1109a23: MOVQ 0x80(SP), AX	                               session.go:434
                   .          .  1109a2b: LEAQ 0x11eee(IP), CX	                            session.go:434
                   .          .  1109a32: MOVQ 0xb8(SP), BX	                               session.go:434
                   .          .  1109a3a: JMP 0x11099a7	                                   session.go:434
                   .          .  1109a3f: LEAQ 0x3d0da(IP), AX	                            session.go:434
```

util.go
```diff
@@ -1,5 +1,20 @@
 package yamux
 
+import (
+	"sync"
+	"time"
+)
+
+var (
+	timerPool = &sync.Pool{
```
erikdubbelboer: This doesn't need to be a pointer but that doesn't really matter.

erikdubbelboer:
I think I found the reason it allocates and your change will still make it allocate:
```go
import (
  "testing"
)

type X struct {
}

func (x *X) switchFunc(a *int) {
}

func BenchmarkSwitch(b *testing.B) {
  x := X{}
  handlers := []func(*X, *int){
    0: (*X).switchFunc,
  }
  for i := 0; i < b.N; i++ {
    a := 1

    x.switchFunc(&a) // Doesn't allocate because it can do proper escape analysis because it knows what switchFunc does.

    h := x.switchFunc
    h(nil) // still doesn't allocate because nil.
    h(&a)  // Allocates because it can't do proper escape analysis for 'a' because it doesn't know what code h executes.

    j := handlers[0]
    j(&x, nil) // Doesn't allocate because nil.
    j(&x, &a)  // Still allocates for the same reason above.
  }
}
```

So the only way to write this to not make it allocate is to write it out fully:
```go
  switch hdr.MsgType() {
  case typeData:
    if err := s.handleStreamMessage(hdr); err != nil {
      return err
    }
  case typeWindowUpdate:
    if err := s.handleStreamMessage(hdr); err != nil {
      return err
    }
  case typeGoAway:          
    if err := s.handleGoAway(hdr); err != nil {
      return err
    }
  case typePing:                        
    if err := s.handlePing(hdr); err != nil {
      return err
    }
  default:                                          
    return ErrInvalidMsgType                            
  }
```

stuartcarnie:
@erikdubbelboer it is not allocating because of a value passed to the function, it is allocating because the original logic was creating a closure, capturing `s`. That is what I demonstrated in the assembly dump. If you look at the assembly for the new code, there is no `newobject` call

```asm
               if err := handlers[mt](s, hdr); err != nil {
  0x1109a80             488b35995a1200          MOVQ 0x125a99(IP), SI
  0x1109a87             488b3d9a5a1200          MOVQ 0x125a9a(IP), DI
// here bounds check of handlers
  0x1109a8e             0fb6db                  MOVZX BL, BX
  0x1109a91             4839fb                  CMPQ DI, BX
// and jump to `panicindex` if CMPQ is BX is outside range
  0x1109a94             0f83e5020000            JAE 0x1109d7f
// find address of function pointer in handlers array for mt; SI contains address of handlers slice
  0x1109a9a             488b14de                MOVQ 0(SI)(BX*8), DX
  0x1109a9e             488b9c2498000000        MOVQ 0x98(SP), BX
// load hdr slice onto stack
  0x1109aa6             48891c24                MOVQ BX, 0(SP)
  0x1109aaa             48894c2408              MOVQ CX, 0x8(SP)
  0x1109aaf             4889442410              MOVQ AX, 0x10(SP)
  0x1109ab4             488b742448              MOVQ 0x48(SP), SI
  0x1109ab9             4889742418              MOVQ SI, 0x18(SP)
// load function pointer of Stream handler into DI
  0x1109abe             488b3a                  MOVQ 0(DX), DI
// call handler
  0x1109ac1             ffd7                    CALL DI
```

erikdubbelboer:
Why would it create a closure? That's not how closures work as far as I know. A closure would be when the code of the function is define there as well so it can use the scope around it.

```go
func a() {
  b := 1
  c := func() { // c is now a closure that captures b.
    println(b)
  }
}
```

I'm 99% sure that your version still allocates because it can't know what the code will do with hdr so it needs to allocate it on the stack. The allocation probably just happens before the dump you posted. The reason it happens inside the `case` in the original code might be because the default case returns immediately and doesn't use hdr.

Try my code with `go test -bench . -benchmem` and uncomment the different variations:
```go
package main

import (
  "testing"
)

var g []byte

type X struct {
}

func (x *X) switchFunc(a []byte) {
  //g = a // Uncommenting this will always make it allocate as it always escapes.
}

func BenchmarkSwitch(b *testing.B) {
  x := &X{}
  handlers := []func(*X, []byte){
    0: (*X).switchFunc,
  }
  _ = handlers
  for i := 0; i < b.N; i++ {
    a := make([]byte, 12)

    x.switchFunc(a) // Doesn't allocate because it can do proper escape analysis because it knows what switchFunc does.

    h := x.switchFunc
    h(nil) // still doesn't allocate because nil.
    //h(a)   // Allocates because it can't do proper escape analysis for 'a' because it doesn't know what code h executes.

    j := handlers[0]
    j(x, nil) // Doesn't allocate because nil.
    //j(x, a)   // Still allocates for the same reason above.

    //handlers[0](x, a) // Still allocates for the same reason above.

    // Doesn't allocate because nil, even though its a closure. The Go compiler is smart.
    k := func() {
      x.switchFunc(nil)
    }
    k()

    // This allocates because it allocates a closure? Or the closure prevents escape analysis?
    //l := func() {
    //  x.switchFunc(a)
    //}
    //l()
  }
}
```

stuartcarnie:
It is a problem with escape analysis. The following will allocate memory in the `methodValue` function. Note the placement of the `handler` variable, outside the `for` loop.

```go
package methods

import (
	"fmt"
	"testing"
	"math/rand"
)

type foo struct {
	s string
}

func (f *foo) handle1(v string) bool {
	fmt.Println(v)
	return true
}

func (f *foo) handle2(v string) bool {
	fmt.Println(f.s + v)
	return true
}

func (f *foo) methodValue(i int) {
	var handler func(string) bool
	for {
		switch i {
		case 0:
			handler = f.handle1

		case 1:
			handler = f.handle2

		default:
			panic("bad")
		}

		handler("hello")
		return
	}
}

func TestFooMethodValue(t *testing.T) {
	f := &foo{}
	f.methodValue(rand.Int() & 1)
}
```

Move the `handler` variable inside the `for` loop, and it no longer allocates.

```go
func (f *foo) methodValue(i int) {
	for {
		var handler func(string) bool
		switch i {
		case 0:
			handler = f.handle1

		case 1:
			handler = f.handle2

		default:
			panic("bad")
		}

		handler("hello")
		return
	}
}
```

stuartcarnie:
running `go build -gcflags '-m'` demonstrates that escape analysis is failing capturing the [Method Value](https://golang.org/ref/spec#Method_values). Note that in the explanation of a method value where I have emphasized what I was explaining about a _copy) of that method value. 

> If the expression x has static type T and M is in the method set of type T, x.M is called a method value. The method value x.M is a function value that is callable with the same arguments as a method call of x.M. The expression x is evaluated and saved during the evaluation of the method value; **the saved copy is then used as the receiver in any calls**, which may be executed later.

Escape analysis failed to determine that this copy did not escape, thus the `newobject` call on these lines.

Here is an output of the escape analysis when the `handler` value is outside the `for` loop, demonstrating the compile is creating a copy of the method value.

```
) go build -gcflags '-m' ./cmd/methods
# github.com/stuartcarnie/play/cmd/methods
... lines removed for brevity
cmd/methods/method.go:22:29: leaking param: f
cmd/methods/method.go:27:15: f.handle1 escapes to heap  <- here capturing the method value of foo
cmd/methods/method.go:30:15: f.handle2 escapes to heap  <- here capturing the method value of foo
cmd/methods/method.go:42:12: &foo literal escapes to heap
... lines removed for brevity
```

after I move the `handler` var into the loop, it no longer escapes

```
) go build -gcflags '-m' ./cmd/methods
# github.com/stuartcarnie/play/cmd/methods
... lines removed for brevity
cmd/methods/method.go:22:29: leaking param: f
cmd/methods/method.go:27:15: (*foo).methodValue f.handle1 does not escape
cmd/methods/method.go:30:15: (*foo).methodValue f.handle2 does not escape
cmd/methods/method.go:42:12: &foo literal escapes to heap
... lines removed for brevity
```

erikdubbelboer:
I see you were right, I was looking at the wrong allocation. Your way indeed saves one allocation, but I still think writing out the full switch is better. Your new solution still suffers from the issue where it can't do proper escape analysis on `hdr` making it always scape. I haven't look at the code to see if other functions make `hdr` escape anyways. But I think having a couple of lines code less is not worth it when it prevent escape analysis from working properly.

stuartcarnie:
We want `hdr` allocated on the heap, so it is reused for each iteration of the loop, to avoid unnecessary allocations :)

erikdubbelboer:
If it's on the stack it can still be reused since the frame for `recvLoop` will stay on the stack. But I guess it doesn't matter that much since it's only one allocation per connection.

session.go
```diff
@@ -323,8 +323,13 @@ func (s *Session) waitForSend(hdr header, body io.Reader) error {
...
 func (s *Session) waitForSendErr(hdr header, body io.Reader, errCh chan error) error {
-	timer := time.NewTimer(s.config.ConnectionWriteTimeout)
-	defer timer.Stop()
+	t := timerPool.Get()
+	timer := t.(*time.Timer)
+	timer.Reset(s.config.ConnectionWriteTimeout)
+	defer func() {
+		timer.Stop()
+		timerPool.Put(t)
+	}()
```
preetapan: seems like sendNoWait could use the same timer pool too?

stream.go
```diff
@@ -238,18 +238,25 @@ func (s *Stream) sendWindowUpdate() error {
 
 	// Determine the delta update
 	max := s.session.config.MaxStreamWindowSize
-	delta := max - atomic.LoadUint32(&s.recvWindow)
```
preetapan: Why did the atomic load go away? I don't have too much context around this code, but seems like the receive lock `recvLock` has a lot more contention since many other places use it, vs previously here it was using atomic load to get a potentially faster lower level primitive to load `recvWindow`.  Can this new logic for bufLen be rewritten to use atomic.loadUint32 for both `recvWindow` and `recvBuf.len()` so we don't need the recvLock mutex?

preetapan:
@stuartcarnie Thanks for the patch. I tried it out, with a small change to the large benchmark to send 512MB rather than 100MB. Pasting results below. Overall, this is looking really good. I'll note that I didn't quite see a 15% speedup, it was more around 8%. Memory usage efficiency gains look really good, it went from allocating 7MB to only needing to allocate 0.3MB, to send 512MB. 

Master branch:
```
preetha@preetha-work ~/go/src/github.com/hashicorp/yamux (master) $go test -bench=. -benchmem
2017/10/04 17:26:55 [WARN] yamux: failed to send ping reply: connection write timeout
goos: linux
goarch: amd64
pkg: github.com/hashicorp/yamux
BenchmarkPing-4            	  200000	      7455 ns/op	     853 B/op	      16 allocs/op
BenchmarkAccept-4          	  100000	     11425 ns/op	    2133 B/op	      33 allocs/op
BenchmarkSendRecv-4        	  500000	      3846 ns/op	     289 B/op	       6 allocs/op
BenchmarkSendRecvLarge-4   	      20	 107960879 ns/op	 7730291 B/op	   20526 allocs/op
PASS
ok  	github.com/hashicorp/yamux	14.524s
```

Your changes:
```
preetha@preetha-work ~/go/src/github.com/hashicorp/yamux (perf) $go test -bench=. -benchmem
2017/10/04 17:28:13 [WARN] yamux: failed to send ping reply: connection write timeout
goos: linux
goarch: amd64
pkg: github.com/hashicorp/yamux
BenchmarkPing-4            	  200000	      6416 ns/op	     624 B/op	      11 allocs/op
BenchmarkAccept-4          	  200000	      9879 ns/op	    1313 B/op	      18 allocs/op
BenchmarkSendRecv-4        	  500000	      3524 ns/op	      80 B/op	       2 allocs/op
BenchmarkSendRecvLarge-4   	      20	  99652741 ns/op	  337417 B/op	    7412 allocs/op
```

I've got a couple of questions in the code review, other than that this is looking good.

stream.go
stuartcarnie: Hi @preetapan, thanks for the feedback.

We must synchronize access to the `recvWindow` and `recvBuf` values, because they are related. The previous (incorrect) logic only looked at the `recvWindow` for determining the value of the window update after a `Read` operation. The bug surfaced when a `Read` does not completely drain the `recvBuf`. We must then take remaining bytes into consideration when calculating the final window update value and therefore take the lock.

session.go
stuartcarnie: Yes, we could use the timer pool for `sendNoWait` too

stuartcarnie:
If you sample the `recvBuf.Cap()` after receiving 100MB, you will see that it is almost 100MB (before the change).  After the change, it is closer to the maximum recv window size.

preetapan:
@stuartcarnie thanks for the explanation. I pushed just the benchmark to master with some debug output. I saw the recvBuf grow up to 16MB when sending 512MB, and yes indeed with your changes it caps out at 2X the window size or 0.5mb.

session.go
```diff
@@ -374,7 +374,7 @@ func (s *Session) send() {
 				for sent < len(ready.Hdr) {
 					n, err := s.conn.Write(ready.Hdr[sent:])
 					if err != nil {
-						s.logger.Printf("[ERR] yamux: Failed to write header: %v", err)
+						s.logger.Printf("[WARN] yamux: Failed to write header: %v", err)
```
erikdubbelboer: Why exactly should it be a warning and not an error? If I understand correctly this will completely stop the yamux connection. I would think with just a warning things would be able to recover and continue to work, not stop completely.

stuartcarnie: @erikdubbelboer none of those additional commits should have appeared in this PR. We are maintaining our own fork.

---

An `error` is still returned to the client, this only changes a _log_ message. I consider yamux consistency issues bugs and should be logged as **errors**. An example of a consistency issue would  `yamux: SYN tracking out of sync` and highlights a bug in yamux. The network connection being closed, causing a write to fail is not a yamux bug and is still reported as an `error` to the client code, however it generates unnecessary log noise.

stuartcarnie:
I am closing this PR as a number of additional commits have unintentionally been pulled in

session.go
erikdubbelboer: In that case should it log anything at all? Maybe it should just return the error and let the application log something if it wants. I always thing libraries logging things if only for when it can't tell its user that something happened (by returning an error for example).

erikdubbelboer:
Are you going to open a new PR? These changes seem very useful.

session.go
stuartcarnie: I definitely agree. The logging within the yamux package should be restricted to yamux internal consistency issues. Logging a warning because the network connection has been closed is redundant, since the error will be passed to the caller

jefferai:
@stuartcarnie There is definite interest in these changes -- it should be easy enough to force push this branch and put the other changes into other PRs. Are you planning on following up?

preetapan:
@jefferai @erikdubbelboer @stuartcarnie I made a new PR that includes the relevant changes (with commit credits to @stuartcarnie ) PR#53

stuartcarnie:
thanks @preetapan

PR #51, opened 2017-10-27
Aboard Read()/Write() by SetDeadline() with past time

yudai:
Hello, thanks for the handy library.

# Problem Observed

When an HTTP server is served on a `yamux.Session`, `Hijack()` on`http.ResponseWriter` blocks forever, which means protocols that depend on hijacking (protocol upgrade) such as Websocket does not work with yamux.

# Problem Detail

In `net/http`, `Set(Read/Write)Deadline()` with [`net.aLongTimeAgo`](https://golang.org/src/net/http/http.go#L23) is used to abort pending `Read()/Write()` by setting the unix epoch + 1 second. `Hijack()` uses this method (and I think there are some more cases) to abort `Read()` as well.

In the document of [`net.Conn`](https://golang.org/pkg/net/#Conn),  `SetDeadline()` is described as below:

> // A deadline is an absolute time after which I/O operations
> // fail with a timeout (see type Error) instead of
> // blocking. The deadline applies to all future and pending
> // I/O, not just the immediately following call to Read or
> // Write. After a deadline has been exceeded, the connection
> // can be refreshed by setting a deadline in the future.

Setting a past time is supposed to cancel any pending `Read()/Write()` on the connection.

However, the current implementation of `yamux.Stream` doesn't support this type of cancelation. Libraries expect this behavior therefore can block forever unexpectedly with yamux.

# Suggested Resolution

When `SetDeadline()` is called, retry `Read()` or `Write()` by using the notification channels.

I think there would be no side effects caused by this change, since when you would not provide a past time to `SetDeadline()` when you actually don't expect cancelation by that.

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -1,7 +1,6 @@
 package yamux
 
 import (
-	"fmt"
 	"testing"
 )
 
@@ -85,7 +84,6 @@ func BenchmarkSendRecvLarge(b *testing.B) {
 	client, server := testClientServer()
 	defer client.Close()
 	defer server.Close()
-
 	const sendSize = 512 * 1024 * 1024
 	const recvSize = 4 * 1024
 
@@ -107,9 +105,6 @@ func BenchmarkSendRecvLarge(b *testing.B) {
 					b.Fatalf("err: %v", err)
 				}
 			}
-
-			fmt.Printf("Capacity of rcv buffer = %v, length of rcv window = %v\n", stream.recvBuf.Cap(), stream.recvWindow)
-
 		}
 		close(recvDone)
 	}()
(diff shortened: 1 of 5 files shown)


itimofeev:
Spent several days because of described problem. When do `hijack()` conn randomly blocked.
Thank you for PR, hope it will be merged soon.

itimofeev:
Figured out some issue with this PR. We also need to return net.Error implementation of ErrTimeout. Or in some cases `httputil.NewSingleHostReverseProxy()` will return `context closed` error. This [commit](https://github.com/axxonsoft/yamux/commit/8e7cea9af99648bdfdb28f9f56947631d07bf6c2) fixes problem.

dadgar:
Closed by https://github.com/hashicorp/yamux/pull/82

PR #52, opened 2017-12-18
Fix race around read and write deadlines in Stream

filipochnik:
I found this race condition when running my code using yamux with `-race`. 

Since the test suite fails when run with `-race` I don't think I can write a meaningful test here.


jefferai:
I think it is likely to be faster to use an `atomic.Value` here, since it is only being read in one place and written in another.

filipochnik:
Sure, I changed it to atomic.Value :)

jefferai:
Thanks!

PR #53, opened 2017-12-19
Streaming perf improv

preetapan:
Created from PR #50

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -238,18 +238,25 @@ func (s *Stream) sendWindowUpdate() error {
 
 	// Determine the delta update
 	max := s.session.config.MaxStreamWindowSize
-	delta := max - atomic.LoadUint32(&s.recvWindow)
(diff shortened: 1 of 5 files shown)


stream.go
```diff
@@ -414,7 +423,7 @@ func (s *Stream) readData(hdr header, flags uint16, conn io.Reader) error {
 	}
 
 	// Decrement the receive window
-	atomic.AddUint32(&s.recvWindow, ^uint32(length-1))
+	s.recvWindow += ^uint32(length - 1)
```
stuartcarnie: Whilst it is the same thing, this is a carry over from the atomic operation. We should change this to read

```
s.recvWindow -= length
```

preetapan:
Did an offline code review with @dadgar to merge this. 

Thanks @stuartcarnie for the great work.

PR #54, opened 2017-12-26
copyright section

aviau:
please mention the original author/copyright holder so we don't have to guess


evanphx:
Strangely I don't see Copyrights in Readmes very often. I'm going to go ahead and close this due to age.

aviau:
What? Can you just specify it anywhere then?

PR #61, opened 2018-06-04
Feat/better reset error

whyrusleeping:
This both makes all yamux errors easier to figure out (took me a while to realize where "connection reset" was coming from) and makes the connection reset error actually say what it means.


whyrusleeping:
whoops. meant to PR against my fork...

PR #66, opened 2018-08-24
Session.recvLoop(): Improve sanity checking of message types.

octo:
Initializing a sparse slice and then only checking upper and lower bounds is a null pointer dereference waiting to happen. Also, explicitly listing the lowest and highest identifier is brittle and prone to break
in the future.


evanphx:
Given the age, I'm going to go ahead and close this PR. The code might be less than ideal but has served the project well thus far.

PR #78, opened 2019-09-09
add RTT in session structure

hanwoody:


evanphx:
This introduces a race condition for setting and reading the RTT. Given the age of the PR, I'm going to go ahead and close it. If you'd like to revise it, please let us know.

PR #79, opened 2020-02-12
add ability to make create streams with timeout like DialTimeout

tantra35:
this issue add ability so partialy solve problem described here https://github.com/hashicorp/nomad/issues/4604, but not fully solve issue


evanphx:
It's unclear why all the header stuff gets changed as well. I like the idea, but it feels like there is a lot more in this PR than just the ability to have a timeout on open.

PR #80, opened 2020-03-23
Add ability to cancel Accept by passing a context

evanphx:


mitchellh:
Looks good, simple enough but we should add tests.

session_test.go
```diff
@@ -1254,3 +1255,30 @@ func TestSession_ConnectionWriteTimeout(t *testing.T) {
...
+
+	var wg sync.WaitGroup
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		stream, err := server.AcceptStreamWithContext(ctx)
+		if err != context.Canceled {
+			t.Fatalf("err: %v", err)
```
notnoop: `t.Fatal` isnt' intended to be called in goroutines (as it calls `t.FatalNow` [1]).  I would suggest setting the error to a variable in function scope and asserting in main goroutine or using a channel to send error instead of using the waitGroup.

[1] https://golang.org/pkg/testing/#T.FailNow

session_test.go
```diff
@@ -1254,3 +1255,30 @@ func TestSession_ConnectionWriteTimeout(t *testing.T) {
...
+		}
+
+		if stream != nil {
+			defer stream.Close()
```
notnoop: nit pick - the differ here isn't necessary I think.

session_test.go
evanphx: I followed the code used elsewhere. If it should be changed, we should do it for all the tests.

session_test.go
evanphx: again, just following the pattern used by the tests.

session.go
```diff
@@ -220,6 +221,24 @@ func (s *Session) AcceptStream() (*Stream, error) {
 	}
 }
 
+// AcceptStream is used to block until the next available stream
+// is ready to be accepted. When the context is canceled, the session
+// is automatically closed.
+func (s *Session) AcceptStreamWithContext(ctx context.Context) (*Stream, error) {
+	select {
+	case <-ctx.Done():
+		s.Close()
```
mitchellh: Do we really want this behavior? It seems like the outer side should have a `defer s.Close` anyways, and I can imagine wanting to perhaps timeout the accept, do some things, and then try again.

PR #82, opened 2020-06-06
Deadline updates impact blocked reads and writes

dadgar:
This change causes new deadlines to impact currently blocked reads and
writes. This is inline with the documentation of deadlines on a
[net.Conn](https://golang.org/pkg/net/#Conn).

This PR can close #https://github.com/hashicorp/yamux/pull/51

### DIFF
--- a/session_test.go
+++ b/session_test.go
@@ -698,6 +698,50 @@ func TestReadDeadline(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


evanphx:
Ah! I see, the "all pending" comment in the net.Conn docs. Excellent!

dadgar:
@evanphx Thanks for the review!

PR #84, opened 2020-09-14
Add support for closing write ends of streams

eandre:
When sending payloads of unknown length over a Stream
and expecting the server to read it to completion before
emitting a response (such as forwarding a byte stream),
one difficulty with yamux is communicating that the byte
stream's end has been reached.

One approach is to introduce a higher-level protocol,
but when the byte stream is of unknown size this requires
essentially reimplementing large parts of yamux's framing
protocol.

A simpler solution is to communicate that EOF has been reached.
Yamux provides this capability through (*Stream).Close,
but it closes both the read and the write ends, which then
prevents the client from reading the response from the server.

This change introduces a new method, (*Stream).CloseWrite,
which only closes the write end of the stream. When encountered
on the other end, it sets a flag that the remote's write end has been
closed and begins returning EOF from any reads after the receive
buffer has been exhausted.

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -21,6 +21,14 @@ const (
 	streamReset
 )
 
+type streamFlags uint16
+
+const (
+	writeCloseFlag streamFlags = 1 << iota
+	writeCloseFlagSent
+	readCloseFlag
+)
+
 // Stream is used to represent a logical stream
 // within a session.
 type Stream struct {
@@ -31,6 +39,7 @@ type Stream struct {
 	session *Session
 
 	state     streamState
+	flags     streamFlags
 	stateLock sync.Mutex
 
 	recvBuf  *bytes.Buffer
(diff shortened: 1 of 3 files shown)


stream.go
```diff
@@ -321,6 +347,53 @@ SEND_CLOSE:
...
+	switch s.state {
+	// Opened means we need to signal a close
+	case streamSYNSent:
+		fallthrough
+	case streamSYNReceived:
+		fallthrough
+	case streamEstablished:
+		goto SEND_CLOSE
+
+	case streamLocalClose:
```
evanphx, line 363: I think you mean that seeing `streamLocalClose` means we do nothing and continue after the switch but Go makes it hard to here if that's the case or if it's a programming error. Could you add a comment to the bottom saying something like "ok, handle normally" or something? Thanks!

evanphx (changes requested):
One comment about the switch, but also can you do a little testing about what happens when a newer yamux sends CloseWrite to an older one that doesn't support the flag? We haven't added new flags to yamux in so many years, it's something we need to consider here.

PR #88, opened 2021-03-15
add StreamCloseTimeout

mitchellh:
This adds a new setting StreamCloseTimeout which forcibly closes the
connection if the remote side doesn't ACK a FIN within the given
timeout. This should only happen in cases the remote side is buggy and
doesn't call `Close` properly.

This allows the Session to release the mapping to the stream and
therefore avoids a memory leak when one side of a stream never completes
a close.

"Forcibly close" in this case means: (1) send a RST (2) delete the
receive buffer (3) delete the stream. Future packets received for the
stream will be dropped (and a log message produced).

### DIFF
--- a/session_test.go
+++ b/session_test.go
@@ -273,6 +273,76 @@ func TestAccept(t *testing.T) {
 	}
 }
 
+func TestClose_closeTimeout(t *testing.T) {
+	conf := testConf()
+	conf.StreamCloseTimeout = 10 * time.Millisecond
+	client, server := testClientServerConfig(conf)
+	defer client.Close()
+	defer server.Close()
+
+	if client.NumStreams() != 0 {
+		t.Fatalf("bad")
+	}
+	if server.NumStreams() != 0 {
+		t.Fatalf("bad")
+	}
+
+	wg := &sync.WaitGroup{}
+	wg.Add(2)
+
+	// Open a stream on the client but only close it on the server.
+	// We want to see if the stream ever gets cleaned up on the client.
+
+	var clientStream *Stream
+	go func() {
+		defer wg.Done()
+		var err error
+		clientStream, err = client.OpenStream()
+		if err != nil {
+			t.Fatalf("err: %v", err)
+		}
+	}()
+
+	go func() {
+		defer wg.Done()
+		stream, err := server.AcceptStream()
+		if err != nil {
+			t.Fatalf("err: %v", err)
+		}
+		if err := stream.Close(); err != nil {
+			t.Fatalf("err: %v", err)
+		}
+	}()
+
+	doneCh := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(doneCh)
+	}()
+
+	select {
+	case <-doneCh:
+	case <-time.After(time.Second):
+		panic("timeout")
+	}
+
+	// We should have zero streams after our timeout period
+	time.Sleep(100 * time.Millisecond)
+
+	if v := server.NumStreams(); v > 0 {
+		t.Fatalf("should have zero streams: %d", v)
+	}
+	if v := client.NumStreams(); v > 0 {
+		t.Fatalf("should have zero streams: %d", v)
+	}
+
+	if _, err := clientStream.Write([]byte("hello")); err == nil {
+		t.Fatal("should error on write")
+	} else if err.Error() != "connection reset" {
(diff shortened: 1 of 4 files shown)


session_test.go
```diff
@@ -273,6 +273,68 @@ func TestAccept(t *testing.T) {
...
+
+	wg := &sync.WaitGroup{}
+	wg.Add(2)
+
+	// Open a stream on the client but only close it on the server.
+	// We want to see if the stream ever gets cleaned up on the client.
+
+	go func() {
+		defer wg.Done()
+		_, err := client.OpenStream()
```
schmichael, line 298: Hm, do Yamux streams close-via-finalizer? If so it seems possible for this test to pass even without your code change as the GC would be implicitly the client-side of the stream.

If this test fails without your patch I think that indicates both this test and the patch are working as intended.

schmichael: Although another way to structure this test might be for the clientside to write in a loop and assert the first error enountered is a `connection reset by peer` and that it occurs after the stream close timeout.

Relying _more_ on timings is rarely a way to make a test more resilient though, so perhaps the added complexity isn't worth it.

stream.go
```diff
@@ -312,6 +316,11 @@ func (s *Stream) Close() error {
 	s.stateLock.Unlock()
 	return nil
 SEND_CLOSE:
+	if !closeStream && s.session.config.StreamCloseTimeout > 0 && s.closeTimer == nil {
+		s.closeTimer = time.AfterFunc(
+			s.session.config.StreamCloseTimeout, s.closeTimeout)
```
schmichael, line 321: This seems to create a goroutine for 5 minutes every time one side closes an established stream. Should we pass a context to `closeTimeout` that gets canceled when `closeStream = true`? It seems like that would remove cleanup goroutines for well behaved peers.

session_test.go
mitchellh: We don't use any finalizers and nothing else will clean a stream up except a session close (which happens way later here). I can be extra sure for the test by retaining a reference though if that helps. :)

Also I did verify the test fails before, I usually do TDD for bugs like this so I had the test written out first.

stream.go
mitchellh: Yeah this was something I missed and I'm adding in now. I don't think we need a context, we can just cancel the timer. 

note timers under the hood don't spawn any goroutines, they sit in a single thread sleeping on a heap.

mitchellh: Pushed a fix for that.

session_test.go
mitchellh: Kept it in scope and also added a write fail test.

session_test.go
```diff
@@ -273,6 +273,74 @@ func TestAccept(t *testing.T) {
...
+	time.Sleep(100 * time.Millisecond)
+
+	if v := server.NumStreams(); v > 0 {
+		t.Fatalf("should have zero streams: %d", v)
+	}
+	if v := client.NumStreams(); v > 0 {
+		t.Fatalf("should have zero streams: %d", v)
+	}
+
+	if _, err := clientStream.Write([]byte("hello")); err == nil {
```
dadgar, line 339: Would be nice to assert the error message is what we expect so if implementation changes over time the error contract stays the same

stream.go
```diff
@@ -312,6 +316,22 @@ func (s *Stream) Close() error {
...
+	// timer is via processFlags) but just in case this ever happens, we
+	// cancel the timer to prevent dangling timers.
+	if s.closeTimer != nil {
+		s.closeTimer.Stop()
+		s.closeTimer = nil
+	}
+
+	// If we have a StreamCloseTimeout set we start the timeout timer.
+	// We do this only if we're not already closing the stream since that
+	// means this was a graceful close.
```
dadgar, line 329: I think I would appreciate as a future reader a example scenario this avoids.

"This prevents a leaked buffer if one side of the stream closes, but the other does not. After the specified timeout, the stream is force closed."

dadgar (approved):

session_test.go
mitchellh: Good call. Done.

stream.go
mitchellh: Good, expanded on the comment.  Done.

schmichael (approved):

mkeeler (approved):

PR #91, opened 2021-07-05
let ErrTimeout implements net.Error

### LINKED ISSUE #90: "context canceled" error when using with http.ReverseProxy
Hi, I'm having this weird problem when using yamux with http.ReverseProxy. The thing I want to achieve is:

1. A server listening on `:8888` for client to connect
2. When a client is connected, the connection is handed over to yamux, and
    - the client becomes the real server, which is just a simple proxy to `example.org`
    - the server starts listening on `:8080`, which would proxy to the client

The weird part is that the first `curl localhost:8080` would success (with 404 from example.org), but then all the following `curl localhost:8080` would return 502 bad gateway, with the client prints out `http: proxy error: context canceled` with each request sent.

Here's some clue:

1. The error message is from the http roundtripper, where it would error out if ctx is done
2. If I avoid yamux, and do server -> client -> example.org proxy directly with http.ReverseProxy, the problem would just go away
3. If I set `DisableKeepAlives: true` to server's transport, the problem would go away too because each request would Dial a new connection then

Below is the detailed code and steps to reproduce the problem:

server code:

```go
package main

import (
	"log"
	"net"
	"net/http"
	"net/http/httputil"
	"net/url"
	"time"

	"github.com/hashicorp/yamux"
)

func main() {
	lis, err := net.Listen("tcp", ":8888")
	if err != nil {
		panic(err)
	}

	conn, err := lis.Accept()
	if err != nil {
		panic(err)
	}

	sess, err := yamux.Server(conn, nil)
	if err != nil {
		panic(err)
	}

	go func() {
		u, err := url.Parse("http://whatever")
		if err != nil {
			panic(err)
		}

		rp := httputil.NewSingleHostReverseProxy(u)
		rp.Transport = &http.Transport{
			Dial: func(network string, addr string) (net.Conn, error) {
				return sess.Open()
			},
		}

		log.Println("serving at :8080")
		if err := http.ListenAndServe(":8080", rp); err != nil {
			panic(err)
		}
	}()

	for {
		time.Sleep(time.Second)
	}
}
```

client code:

```golang
package main

import (
	"log"
	"net"
	"net/http"
	"net/http/httputil"
	"net/url"

	"github.com/hashicorp/yamux"
)

func main() {
	conn, err := net.Dial("tcp", "localhost:8888")
	if err != nil {
		panic(err)
	}

	sess, err := yamux.Client(conn, nil)
	if err != nil {
		panic(err)
	}

	u, err := url.Parse("http://example.org")
	if err != nil {
		panic(err)
	}

	log.Println("serving")
	if err := http.Serve(sess, httputil.NewSingleHostReverseProxy(u)); err != nil {
		panic(err)
	}
}
```

And do below in terminal to see the problem:

```shell
curl localhost:8080 // success
curl localhost:8080 // 502 bad gateway
```

fatedier:
Fix #90 

Golang standard http server package will read from connections in background to detect if it's alive. It set read deadline on connection and detect if the returned error is timeout error which implements net.Error.

If not, it will call `cr.handleReadError(err)` and then call `cr.conn.cancelCtx()`, so server `request.Context()` will be canceled too.

```go
func (cr *connReader) backgroundRead() {
    n, err := cr.conn.rwc.Read(cr.byteBuf[:])
    cr.lock()
    if n == 1 {
        cr.hasByte = true
        // We were past the end of the previous request's body already
        // (since we wouldn't be in a background read otherwise), so
        // this is a pipelined HTTP request. Prior to Go 1.11 we used to
        // send on the CloseNotify channel and cancel the context here,
        // but the behavior was documented as only "may", and we only
        // did that because that's how CloseNotify accidentally behaved
        // in very early Go releases prior to context support. Once we
        // added context support, people used a Handler's
        // Request.Context() and passed it along. Having that context
        // cancel on pipelined HTTP requests caused problems.
        // Fortunately, almost nothing uses HTTP/1.x pipelining.
        // Unfortunately, apt-get does, or sometimes does.
        // New Go 1.11 behavior: don't fire CloseNotify or cancel
        // contexts on pipelined requests. Shouldn't affect people, but
        // fixes cases like Issue 23921. This does mean that a client
        // closing their TCP connection after sending a pipelined
        // request won't cancel the context, but we'll catch that on any
        // write failure (in checkConnErrorWriter.Write).
        // If the server never writes, yes, there are still contrived
        // server & client behaviors where this fails to ever cancel the
        // context, but that's kinda why HTTP/1.x pipelining died
        // anyway.
    }
    if ne, ok := err.(net.Error); ok && cr.aborted && ne.Timeout() {
        // Ignore this error. It's the expected error from
        // another goroutine calling abortPendingRead.
    } else if err != nil {
        cr.handleReadError(err)
    }
    cr.aborted = false
    cr.inRead = false
    cr.unlock()
    cr.cond.Broadcast()
}
```

We'd better let ErrTimeout defined in consts.go to satisfy `net.Error` interface, so everything will be ok.

```go
// An Error represents a network error.
type Error interface {
    error
    Timeout() bool   // Is the error a timeout?
    Temporary() bool // Is the error temporary?
}
```

cc @mitchellh @dadgar

### DIFF
--- a/const.go
+++ b/const.go
@@ -5,6 +5,25 @@ import (
 	"fmt"
 )
 
+// NetError implements net.Error
+type NetError struct {
+	err       error
+	timeout   bool
+	temporary bool
+}
+
+func (e *NetError) Error() string {
+	return e.err.Error()
+}
+
+func (e *NetError) Timeout() bool {
+	return e.timeout
+}
+
+func (e *NetError) Temporary() bool {
+	return e.temporary
+}
+
 var (
 	// ErrInvalidVersion means we received a frame with an
 	// invalid version
@@ -30,7 +49,10 @@ var (
 	ErrRecvWindowExceeded = fmt.Errorf("recv window exceeded")
 
 	// ErrTimeout is used when we reach an IO deadline
-	ErrTimeout = fmt.Errorf("i/o deadline reached")
+	ErrTimeout = &NetError{
+		err:     fmt.Errorf("i/o deadline reached"),
+		timeout: true,
+	}
 
 	// ErrStreamClosed is returned when using a closed stream
 	ErrStreamClosed = fmt.Errorf("stream closed")
--- a/session_test.go
+++ b/session_test.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"net"
 	"reflect"
 	"runtime"
 	"strings"
@@ -763,8 +764,16 @@ func TestReadDeadline(t *testing.T) {
 	}
 
 	buf := make([]byte, 4)
-	if _, err := stream.Read(buf); err != ErrTimeout {
+	_, err = stream.Read(buf)
+	if err != ErrTimeout {
 		t.Fatalf("err: %v", err)
+	}
+	// See https://github.com/hashicorp/yamux/issues/90
+	// Standard http server package will read from connections in background to detect if it's alive.
+	// It sets read deadline on connections and detect if the returned error is timeout error which implements net.Error.
+	// HTTP server will cancel all server requests if it isn't timeout error from connections.
+	if netErr, ok := err.(net.Error); !ok || !netErr.Timeout() {
+		t.Fatalf("error of reading timeout is expected to implement net.Error and return true when calling Timeout(), but not")
 	}
 }
 


dadgar (commented):
This makes sense to me. Can we add a test that has a timeout and assert that the returned error meets the network Error interface and marks it as a timeout. In the comment of that test can you describe the reason we need this.

fatedier:
> This makes sense to me. Can we add a test that has a timeout and assert that the returned error meets the network Error interface and marks it as a timeout. In the comment of that test can you describe the reason we need this.

OK. I will add a test.

dadgar (approved):

dadgar:
Thanks for the contribution @fatedier

PR #96, opened 2021-08-24
Implement timeouts for OpenStream

freddygv:
Currently streams can be in the streamSYNSent state for an unbounded
amount of time when the remote is unable to reply and has not closed the
connection via FIN/RST.

This commit adds a deadline to receiving an ACK after a SYN is sent on
OpenStream.

The timeout is opt-in only and configurable per-session with StreamOpenTimeout.
When the timeout is triggered, an error is logged.

Note that the establishCh is buffered and we avoid closing the channel because
there are multiple call-paths that can invoke it: 
- Stream establishment
- Remote close (FIN/RST)
- Local close

### DIFF
--- a/mux.go
+++ b/mux.go
@@ -31,6 +31,11 @@ type Config struct {
 	// window size that we allow for a stream.
 	MaxStreamWindowSize uint32
 
+	// StreamOpenTimeout is the maximum amount of time that a stream will
+	// be allowed to remain open until it is ACKed.
+	// Once the timeout is reached the session will be gracefully closed.
+	StreamOpenTimeout time.Duration
+
 	// StreamCloseTimeout is the maximum time that a stream will allowed to
 	// be in a half-closed state when `Close` is called before forcibly
 	// closing the connection. Forcibly closed connections will empty the
@@ -56,6 +61,7 @@ func DefaultConfig() *Config {
 		ConnectionWriteTimeout: 10 * time.Second,
 		MaxStreamWindowSize:    initialStreamWindow,
 		StreamCloseTimeout:     5 * time.Minute,
+		StreamOpenTimeout:      30 * time.Second,
 		LogOutput:              os.Stderr,
 	}
 }
--- a/session.go
+++ b/session.go
@@ -184,6 +184,10 @@ GET_ID:
 	s.inflight[id] = struct{}{}
 	s.streamLock.Unlock()
 
+	if s.config.StreamOpenTimeout > 0 {
+		go s.setOpenTimeout(stream)
+	}
+
 	// Send the window update to create
 	if err := stream.sendWindowUpdate(); err != nil {
 		select {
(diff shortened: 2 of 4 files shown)


banks:
> The timeout is opt-in only and configurable per-session with StreamOpenTimeout.

I'd maybe lean towards setting it to a reasonable default and make it opt-out. This is slightly less risk averse but still gives a path in the unlikely event this causes problems. I can't really think though of cases where OpenStream is blocking for more than say 30 seconds where just stalling indefinitely is actually good behaviour. It is possible that there are cases where it's not harmful and this change will cause more session closes which might introduce more visible errors in other in-flight streams, but I don't think it's especially dangerous and having a default configuration that avoids this failure mode seems better to me.

What do you think?

mux.go
```diff
@@ -31,6 +31,11 @@ type Config struct {
 	// window size that we allow for a stream.
 	MaxStreamWindowSize uint32
 
+	// StreamOpenTimeout is the maximum amount of time that a stream will
+	// be allowed to remain open until it is ACKed.
```
banks, line 35:
```suggestion
	// be allowed to remain in pending state while waiting for an ack from the peer.
```

mux.go
```diff
@@ -31,6 +31,11 @@ type Config struct {
...
+	// Once the timeout is reached the session will be gracefully closed.
```
banks, line 36:
```suggestion
	// Once the timeout is reached the session will be gracefully closed.
	// A zero value disables the StreamOpenTimeout allowing unbounded
	// blocking on OpenStream calls.
```

session.go
```diff
@@ -196,6 +200,27 @@ GET_ID:
 	return stream, nil
 }
 
+// setOpenTimeout implements a timeout for streams that are opened but not established.
+// If the StreamOpenTimeout is exceeded we assume the server is unable to ACK,
```
banks, line 204:
```suggestion
// If the StreamOpenTimeout is exceeded we assume the peer is unable to ACK,
```
Yamux can open streams in either direction not just client to server in the TCP sense.

banks (approved):
LGTM.

I posted a question above about whether we should make the timeout default config - I'd lean towards yes I think provided it can be opted out as the current indefinite blocking seems like a bug and this is a reasonable fix. The risk of disruption seems low given that something has to be pretty wrong with the peer or network to prevent it from acking for so long anyway.

freddygv:
@banks Yeah that was my initial thought by adding a default of 30s. I think that setting a reasonable default and allowing it to be disabled seems conservative enough.

I did some sleuthing to see what defaults exist for this similar settings in the wild and am seeing 75s on both FreeBSD and Mac OS.

See `TCP_KEEPINIT` here:
https://www.freebsd.org/cgi/man.cgi?query=tcp&apropos=0&sektion=4&manpath=FreeBSD+9.2-RELEASE&arch=default&format=html 

How does that sound?

Issue #40, opened 2016-10-10
Send timeout might corrupt outgoing messages


Matthias247:
Hi everyone,

I just reviewed the library a little bit since I'm also interested in stream multiplexing (actually want something for websockets instead of plain TCP, but I guess yamux would work there too).
Overall it looks really well written and I like the fact that you have a spec :-)

One thing that I stumbled upon during review was there might be situations where the send timeout for the session might corrupt outgoing data: If we look at the `waitForSendErr` function it will start the timout and try to send the data in parallel be handing it over to the send channel. 

The select/timeout logic looks like it introduces a data race to me: If the timeout is expiring here the send will still be in progress. That means the data (header and body) will still be in use in the send routine. However the `waitForSendErr` call returns with an the calling stream could think the data is no longer required and reuse it. It might probably not do it for `sendHdr` as it sees that writes fail and no further writes and `sendHdr` mutations follow, but the payload data slice might be reused by the application and mutated - while the send is still in progress. As an end result you could send data to the wire that was not desired.

Fixing this is probably not that easy. I guess that there probably could be some cancellation or sendInProgress flags in the enqueued `sendReady` struct. If the message was not yet fetched by the writing proc it can be marked as cancelled and will there no longer be processed. However as soon as the writer goroutine catches it the safest thing is to let it write to completion as there is no kind of CancellationToken that can be passed to to the `Write` or `Copy` methods.

dosgo:
I also encountered this problem!

mafredri:
I believe I've identified the same underlying issue, which I documented in https://github.com/hashicorp/yamux/pull/100. AFAICT, `sendHdr` is not guaranteed to be protected on access, especially considering `sendCh` is buffered.

PR #102, opened 2022-07-11
Protect reads of `sendReady.Body` with mutex and temp buffer

mafredri:
This commit fixes an edge case where the `body` passed to `waitForSendErr` can be written to after returning from the function. This happens when `sendReady` is buffered on the `sendCh` and the session is shutdown or the write times out.

When this condition happens and `waitForSendErr` has not yet exited, the `body` is safely copied into a temporary buffer in `send`. Otherwise `waitForSendErr` safely created a copy of the `body` and exits, this essentially results in double buffering for the edge case which seems acceptable.

Related to: #40
Detailed analysis: https://github.com/coder/coder/issues/2429

### DIFF
--- a/session.go
+++ b/session.go
@@ -2,6 +2,7 @@ package yamux
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -80,6 +81,7 @@ type Session struct {
 // or to directly send a header
 type sendReady struct {
 	Hdr  []byte
+	mu   *sync.Mutex // Protects Body from unsafe reads.
 	Body []byte
 	Err  chan error
 }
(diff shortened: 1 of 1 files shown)


session.go
```diff
@@ -382,12 +384,34 @@ func (s *Session) waitForSendErr(hdr header, body []byte, errCh chan error) erro
...
+		// underlying array.
+		ready.mu.Lock()
+		defer ready.mu.Unlock()
+
+		if ready.Body == nil {
+			return // Body was already copied in `send`.
+		}
+		newBody := make([]byte, len(body))
+		copy(newBody, body)
+		ready.Body = newBody
```
mafredri, line 404: An alternative to creating this copy here would be to invalidate the `sendReady` request as long as the header hasn't been written. I implemented it like this to keep the behavior as close to current as possible. If changed, `ready.Body` should be copied earlier in `send`, after which `sendReady` no longer can be invalidated.

jefferai:
Thank you!

PR #103, opened 2022-07-11
Fix unsafe header re-use on close and timeout

### LINKED ISSUE #40: Send timeout might corrupt outgoing messages

mafredri:
This commit takes a minimal approach to fixing unsafe header re-use during close and timeout by reallocating the header in those cases.

Fixes #40
Related #102

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -2,6 +2,7 @@ package yamux
 
 import (
 	"bytes"
+	"errors"
 	"io"
 	"sync"
 	"sync/atomic"
@@ -200,6 +201,10 @@ START:
 	// Send the header
 	s.sendHdr.encode(typeData, flags, s.id, max)
 	if err = s.session.waitForSendErr(s.sendHdr, body, s.sendErr); err != nil {
+		if errors.Is(err, ErrSessionShutdown) || errors.Is(err, ErrConnectionWriteTimeout) {
+			// Message left in ready queue, header re-use is unsafe.
+			s.sendHdr = header(make([]byte, headerSize))
+		}
 		return 0, err
 	}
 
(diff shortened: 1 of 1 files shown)


stream.go
```diff
@@ -200,6 +201,10 @@ START:
 	// Send the header
 	s.sendHdr.encode(typeData, flags, s.id, max)
 	if err = s.session.waitForSendErr(s.sendHdr, body, s.sendErr); err != nil {
+		if errors.Is(err, ErrSessionShutdown) || errors.Is(err, ErrConnectionWriteTimeout) {
```
evanphx, line 204: This makes yamux depend on Go 1.13 or later. That's not necessarily a problem, but it should be noted as such.

jefferai: I don't even think it needs to be noted, TBH. Go 1.13 is _well_ outside the security update window and should not be used any longer (or for the past year +)

PR #105, opened 2022-07-13
wait for both recv and send routines to complete on close

clems4ever:
this change ensure that there is no write operations after close is
called. This was not the case before and this led to race conditions
in the caller code because there was no way to ensure there would not be
any write anymore after the close.


jefferai:
Looks good -- can you fix the conflicts? Thanks!

jefferai:
I did a local conflict resolution but after doing so I'm seeing a failing test.

jefferai:
The failing test actually turned out to be the same underlying problem as #77 !

I've updated the test and it passes.

jefferai:
Looks good, thanks!

clems4ever:
Awesome thanks!

PR #106, opened 2022-07-18
fix: Make sendReady a pointer so the Body can be swapped

mafredri:
This pr re-fixes the issue fix attempted in #102.

Due to a refactor of the implementation, the non-pointer characteristic was overlooked and we were modifying a copy of `sendReady`, this fixes that.


jefferai:
Thanks!

PR #108, opened 2022-07-19
Add AcceptBacklog function

manish-hashicorp:
This will return the number of streams ready to be accepted. It can be useful to export this value as a metric.


evanphx (approved):
Exposing this is fine with me.

PR #121, opened 2023-06-28
let ErrTimeout temporary

fatedier:
`ErrTimeout` should be temporary, just like the standard packages in Golang.

```go
// Read header, payload.
if err := c.readFromUntil(c.conn, recordHeaderLen); err != nil {
    // RFC 8446, Section 6.1 suggests that EOF without an alertCloseNotify
    // is an error, but popular web sites seem to do this, so we accept it
    // if and only if at the record boundary.
    if err == io.ErrUnexpectedEOF && c.rawInput.Len() == 0 {
        err = io.EOF
    }
    if e, ok := err.(net.Error); !ok || !e.Temporary() {
        c.in.setErrorLocked(err)
    }
    return err
}
```

When we use tls.Conn to wrap yamux.Stream, if a read deadline occurs, it will cause subsequent read operations to fail continuously because `e.Temporary()` returns false, indicating a permanent error.

@jefferai PTAL

### DIFF
--- a/const.go
+++ b/const.go
@@ -54,7 +54,8 @@ var (
 
 		// Error should meet net.Error interface for timeouts for compatability
 		// with standard library expectations, such as http servers.
-		timeout: true,
+		timeout:   true,
+		temporary: true,
 	}
 
 	// ErrStreamClosed is returned when using a closed stream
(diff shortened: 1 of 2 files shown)


prashil-g:
@tgross Can you help merge this PR
@fatedier FRPS uses very old version on this pinned commit. should we rebase and pin the newer commit if merge is taking time?

PR #130, opened 2024-08-05
chore: update tests to pass the race detector and linter

rboyer:
It now passes:

- `golangci-lint run ./...`
- `go test -race ./... -short`

The `-short` is needed because both of `TestSendData_VeryLarge` and `TestSendData_Large` fail for timing reasons (since race mode is generally slower) and I was hesitant to rewrite them to be less timing sensitive, so I have them get skipped when in testing short mode for now.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,41 @@
+name: CI Tests
+on:
+  pull_request:
+    paths-ignore:
+      - 'README.md'
+  push:
+    branches:
+      - 'master'
+    paths-ignore:
+      - 'README.md'
+
+permissions:
+  contents: read
+
+jobs:
+  go-fmt-and-vet:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+    - uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+      with:
+        go-version: '1.20'
+        cache: true
+    - run: |
+        files=$(go fmt ./...)
+        if [ -n "$files" ]; then
+          echo "The following file(s) do not conform to go fmt:"
+          echo "$files"
+          exit 1
+        fi
+  go-test:
+    needs: go-fmt-and-vet
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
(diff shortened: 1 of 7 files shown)


session_test.go
```diff
@@ -379,107 +343,110 @@ func TestClose_closeTimeout(t *testing.T) {
...
 	if err != nil && conn != nil {
-		t.Error("bad: accept should return a connection of nil value")
+		t.Fatal("bad: accept should return a connection of nil value")
 	}
 
 	conn, err = server.Open()
-	if err != nil && conn != nil {
-		t.Error("bad: open should return a connection of nil value")
+	if err == nil || !errors.Is(err, ErrSessionShutdown) || conn != nil {
+		t.Fatal("bad: accept should return a shutdown error and a connection of nil value")
```
schmichael, line 359:
```suggestion
		t.Fatal("bad: open should return a shutdown error and a connection of nil value")
```

session_test.go
```diff
@@ -379,107 +343,110 @@ func TestClose_closeTimeout(t *testing.T) {
...
+	_, server := testClientServer(t)
 	server.Close()
 
 	conn, err := server.Accept()
+	if err == nil || !errors.Is(err, ErrSessionShutdown) || conn != nil {
+		t.Fatal("bad: accept should return a shutdown error and a connection of nil value")
+	}
 	if err != nil && conn != nil {
-		t.Error("bad: accept should return a connection of nil value")
+		t.Fatal("bad: accept should return a connection of nil value")
```
schmichael, line 354: nbd but match original code and test both cases in one fell swoop

```suggestion
		t.Error("bad: accept should return a connection of nil value")
```

schmichael (approved):

.github/workflows/test.yaml
```diff
@@ -0,0 +1,41 @@
...
+  contents: read
+
+jobs:
+  go-fmt-and-vet:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
+    - uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+      with:
+        go-version: '1.20'
```
tgross, line 22: Not a more recent version?

tgross (approved):
LGTM!

.github/workflows/test.yaml
schmichael: I just copied and pasted [raft's](https://github.com/hashicorp/raft/blob/main/.github/workflows/ci.yml) as a basis assuming they're the best maintained shared library. :grimacing:  We can bump it. I'm not really sure what the optimal strategy is for a little library like this with no deps and no-post-1.14 features afaik.

PR #131, opened 2024-08-14
Allow half-closed reads and test against TCP/TLS connections

schmichael:
**Summary**

1. Switched many tests to use TLS/TCP connections instead of just io.Pipes.
2. Fixed a bug where reads on half-closed streams would fail.

**Details**

In an attempt to debug hashicorp/nomad#23305 I switched tests from using `io.Pipe` to `tls.Conn` for transports.

While this did not uncover the root cause of 23305, it did catch a case where the implementation did not follow the yamux or SPDY specs: after calling Stream.Close, you can still call Stream.Read. Reads on half-closed streams had been explicitly disallowed in 912e29616f9b41d9cbe0d35bc76bd43ef6f11943 for unclear reasons. Even at the time `TestHalfClose` was unaffected by this change. Switching the test from io.Pipe to tls.Conn uncovered the bug. See https://www.chromium.org/spdy/spdy-protocol/spdy-protocol-draft3-1/#236-stream-half-close for details.

Individual commits have more details and are hopefully grouped reasonably. In particular 5bda672f5865826e0b27d534c812f69ab39ba239 describes the new test helpers.

**#127**

Sadly no test results changed before or after #127 was applied. This is not entirely surprising as we already suspected tests did not cover the changes being made.


session_test.go
```diff
@@ -67,14 +68,152 @@ func (p *pipeConn) Close() error {
...
+		t.Fatalf("error loading certificate: %v", err)
+	}
+
+	l, err := net.ListenTCP("tcp", nil)
+	if err != nil {
+		t.Fatalf("error creating listener: %v", err)
+	}
+	t.Cleanup(func() { _ = l.Close() })
+
+	var srv net.Conn
```
rboyer, line 128: This feels slightly racy. Probably better to send this `net.Conn` back out of the goroutine using another channel?

session_test.go
```diff
@@ -67,14 +68,152 @@ func (p *pipeConn) Close() error {
...
+	l, err := net.ListenTCP("tcp", nil)
+	if err != nil {
+		t.Fatalf("error creating listener: %v", err)
+	}
+	t.Cleanup(func() { _ = l.Close() })
+
+	network := l.Addr().Network()
+	addr := l.Addr().String()
+
+	var server net.Conn
```
rboyer, line 89: This feels slightly racy. Probably better to send this `net.Conn` back out of the goroutine using another channel?

schmichael: `server` is only used after receiving on `errCh` so errCh synchronizes it. (CI also runs with `-race`, and I did things like `go test -race -count=1000` locally when trying desperately to find a root cause of what was seen in #127.)

session_test.go
schmichael: Same as above, errCh synchronizes access to srv... although switching naming schemes is a bit annoying. I'll fix that up.

tgross (approved):
LGTM! :+1:

session_test.go
rboyer: Oh man I looked right over `defer close(errCh)` while skimming for `errCh <- nil` :facepalm:

rboyer (approved):

Issue #139, opened 2024-12-01
Race Condition on Concurrent Reads


lukaspj:
Hello!

The docs have recently been updated to state that `Read` can be called concurrently for Streams:
https://github.com/hashicorp/yamux/pull/136

However, I think that this is not the case.

We are using a Goroutine reading an empty byte array to detect remote stream closure like:

```go
  go func() {
    for {
      _, err := stream.Read([]byte{})
      if err != nil {
        cancel()
        _ = stream.SetDeadline(time.Now())
        _ = stream.Close()
        break
      }
    }
  }()
```
We see this read and our primary read going into a deadlock because if the remote closes then this empty read will first run `s.stateLock.Lock()`:
https://github.com/hashicorp/yamux/blob/master/stream.go#L102

And then `s.recvLock.Lock()`:
https://github.com/hashicorp/yamux/blob/master/stream.go#L109

But the other Read is in the midst of sendWindowUpdate:
https://github.com/hashicorp/yamux/blob/master/stream.go#L134

So it has called `s.recvLock.Lock()`:
https://github.com/hashicorp/yamux/blob/master/stream.go#L123

Then it went into the `sendFlags` function:
https://github.com/hashicorp/yamux/blob/master/stream.go#L285

And it is now trying to run `s.stateLock.Lock()`:
https://github.com/hashicorp/yamux/blob/master/stream.go#L255

The difference in order of locking causes stateLock to be locked in thread 1 whereas recvLock is locked in thread 2 and each thread wants both of the locks.

tgross:
Hi @lukaspj! I was the one who made that docs change and you're right that it's not correct. The context of the change was that someone noted that you'd get inconsistent reads when trying to read from a stream from different goroutines, but the `net.Conn` interface in the stdlib (which `Stream` allegedly implements) says you can. I confess I failed to investigate whether it was _actually_ safe to call these method concurrently and only patched the docs to cover not doing what that person reported.

But it's clear that's incorrect. Unfortunately there's no way to express "is safe to use concurrently" in the Go type system, which makes the doc comments on `net.Conn` not strict enough to enforce we're giving you a `net.Conn` that follows those semantics exactly. At this point, it would be a major breaking change for us to do so anyways.

I'm going to re-fix the documentation to make it clear that you should only service a `Stream` from a single goroutine; this is how all the HashiCorp consumers do this already.

[...]

PR #140, opened 2024-12-03
docs: stream Reads are not safe for concurrent use

tgross:
In #136 we documented warnings around the semantics of calling `Stream.Read` concurrently. But we assumed without investigating that the concurrent operations were actually safe. But in fact the locks made by multiple Read calls can interleave, causing a deadlock. Update the documentation again to make this clear.

Fixes: https://github.com/hashicorp/yamux/issues/139

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -22,8 +22,10 @@ const (
 	streamReset
 )
 
-// Stream is used to represent a logical stream
-// within a session.
+// Stream is used to represent a logical stream within a session. Methods on
+// Stream are safe to call concurrently with one another, but all Read calls
+// must be on the same goroutine and all Write calls must be on the same
+// goroutine.
 type Stream struct {
 	recvWindow uint32
 	sendWindow uint32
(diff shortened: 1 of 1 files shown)


lukaspj:
Do you think that stream closure can still be safely detected by reading an empty array like this example?
https://github.com/hashicorp/yamux/pull/115#pullrequestreview-2346749216

Or do we need another pattern?

schmichael:
> Do you think that stream closure can still be safely detected by reading an empty array like this example? [#115 (review)](https://github.com/hashicorp/yamux/pull/115#pullrequestreview-2346749216)
> 
> Or do we need another pattern?

That pattern is fine as long as you do not expect data or concurrent Reads on that Stream. At least until we fix the deadlock there should only be 1 Reader and 1 Writer per Stream.

That being said, even if we fix the deadlock I can not think of a case where multiple Readers on a single Stream is coherent or at least optimal. Multiple concurrent Readers would receive different chunks of Writes to the Stream, and lose the order they came in which would prevent reassembly. I cannot think of a way that multiple concurrent Readers (with buffers) could produce coherent results.

Your use case of splitting "main" Reads and EOF detection into 2 goroutines doesn't run into coherency issues, but it still isn't optimal as on data the empty-buffer Reader will spin in a tight loop until the main Reader actually Reads the buffer. In practice you may only see 0-2 empty-buffer Reads before the main Reader picks up the data. However, I would suspect in loaded conditions where the main Reader goroutine may spend a significant amount of time performing operations on received data before calling Read again: the empty-buffer Reader may spin quite a lot and waste considerable CPU time.

So my recommended pattern, regardless of deadlock/race-condition bugs, is to always have at most 1 Reader and 1 Writer goroutine for yamux Streams (any goroutine can call Close). Your main Reader should already be detecting and handling EOFs, so hopefully removing the empty-buffer goroutine doesn't complicate your code. You can always keep your EOF-detection goroutine, give it a buffer, and on data copy the buffer over a channel or pipe to the main Reader instead of having the main Reader use the Stream directly.

schmichael (approved):

lukaspj:
Ah yeah not an issue for us at all, just need to know what we can and can’t do then we’re good 😊
Thanks a lot! Really appreciate all the time you put into this!

mukeshjc (approved):
LGTM

PR #143, opened 2025-01-14
Close Session when there's failures to write to a Stream

### LINKED ISSUE #142: DefaultConfig has dangerous defaults causing hung Reads
The default values for `Session.config.KeepAliveInterval` and `Session.config.ConnectionWriteTimeout` of `30s` and `10s` create the possibility for timed out writes that most (Hashicorp Nomad included) aren't handling in their readers.

Calls to `Stream.Read` on one side of a connection will hang until the underlying `Session` is closed if the corresponding `Stream.Write` call on the other side it's waiting for returns with `ErrConnectionWriteTimeout`. This happens in the case of network congestion between the two sides.

If you keep `Session.sendCh` full (fixed capacity of 64) for `ConnectionWriteTimeout`, but for less than the `KeepAliveInterval + ConnectionWriteTimeout` (which would kill the `Session`), `Stream.Write` will return `ErrConnectionWriteTimeout`. The state of the underlying `Session` or `Stream` is not modified. When this happens (or doesn't, heh), the other side's `Stream.Read` call that's waiting for that write will never return because there's no timeout for this edge-case.

Since no keep alive timed out, you can continue to use the `Session` once the network congestion is resolved, but that `Stream.Read` call will only return when the `Session` closes or the response shows up. Since the write call on the other side timed out... well, that's a problem.

I can see three possible fixes, one heavier handed than the other-
1. If `Stream.Write` times out, it should implicitly call `Stream.Close` while returning, to notify any blocked calls to `Stream.Read` on the other side return.
    * If the call to `Stream.Close` fails then `Session.Close` should get called, which it doesn't look like that happens now either.
2. A call to `Stream.Write` that times out should just close the entire `Session`.
3. (The hacky fix) Make the default keep alive interval less than the write timeout

lattwood:
This pull request introduces a new test to detect a deadlock bug and includes modifications to handle connection write timeouts more robustly. The changes span across multiple files, with the primary focus on ensuring that sessions are closed properly when a write timeout occurs.

### New Test for Deadlock Detection:
* [`deadlock_test.go`](diffhunk://#diff-1a6b4d7b915c64d57fdd91cf8496c005beca0b741f24a138cfa7498b81ee5904R1-R266): Added a comprehensive test `TestTimeoutParallel` to simulate and detect a deadlock scenario in yamux streams. This test sets up two streams, stalls one to fill the channel, and verifies if the session closes correctly on write timeout.

### Improvements in Connection Write Timeout Handling:
* `session.go`: Enhanced the `waitForSendErr` and `sendNoWait` functions to ensure that the session is properly closed when a connection write timeout or session shutdown occurs. The changes include:
  * Adding deferred functions to call `s.exitErr(err)` if an error occurs. [[1]](diffhunk://#diff-3d4a6b7d54b5107bc5694b7a06383f9ec4c68f9abe43ba0942f5d970f3a8ccacL391-R391) [[2]](diffhunk://#diff-3d4a6b7d54b5107bc5694b7a06383f9ec4c68f9abe43ba0942f5d970f3a8ccacR404-R418) [[3]](diffhunk://#diff-3d4a6b7d54b5107bc5694b7a06383f9ec4c68f9abe43ba0942f5d970f3a8ccacL434-R456) [[4]](diffhunk://#diff-3d4a6b7d54b5107bc5694b7a06383f9ec4c68f9abe43ba0942f5d970f3a8ccacR469-R482)

### Test Adjustments:
* [`session_test.go`](diffhunk://#diff-f27c62978dcbd8717b591269e831e3cc0eec4fbdb8fbf53f66435cd8450c4557R1666-R1671): Modified the `TestSession_PingOfDeath` to handle the expected `ErrSessionShutdown` error when a write times out, ensuring the test accounts for the new session shutdown behaviour.

Closes #142.

I humbly request a callout in changelogs when this bugfix ships :)

### DIFF
--- a/deadlock_test.go
+++ b/deadlock_test.go
@@ -126,7 +126,6 @@ func TestTimeoutParallel(t *testing.T) {
 	var wg sync.WaitGroup
 	// used for failing now
 	errCh := make(chan error, 1)
-	t.Parallel()
 
 	// 1. Setup net.Pipe
 	serverSide, clientSide := makeConnPair()
--- a/deadlock_test.go
+++ b/deadlock_test.go
@@ -1,0 +1,267 @@
+// deadlock_test.go
+//
+// This test sets up two yamux streams in parallel (A and B). We artificially
+// stall the underlying pipeConn so that B's writes fill yamux's channel
+// and eventually time out, but we do NOT close the session. This leaves the
+// remote side's B.Read call blocked forever, illustrating the deadlock bug.
+//
+// Meanwhile, stream A is paused and resumed just to illustrate concurrency,
+// and that the other streams are unaffected- but it's not strictly
+// necessary. The main point is that B times out on write and does *not*
+// close the session, so the remote Read is stuck forever.
+
+package yamux
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"sync"
+	"testing"
+	"time"
+)
+
+// We'll use smaller timeouts so the scenario triggers quickly.
+const (
+	testConnWriteTimeout = 300 * time.Millisecond
+	testKeepAlive        = 30 * time.Second // KeepAliveInterval > ConnectionWriteTimeout
+)
+
+type msg struct {
+	label string
+	val   int
+}
+
+// monotonicWriter sends a never-ending sequence of json objects with incrementing
+// integers, to the provided stream. It optionally pauses when signaled on the pausedCh.
+func monotonicWriter(t *testing.T, stream io.WriteCloser, label string, pausedCh <-chan struct{}) error {
+	var counter int
+	enc := json.NewEncoder(stream)
+	for {
+		// Optionally pause if channel is signaled
+		select {
+		case <-pausedCh:
+			// wait until resumed
+			t.Logf("[%s] writer paused", label)
+			<-pausedCh
+			t.Logf("[%s] writer resumed", label)
+		default:
+		}
+
+		err := enc.Encode(msg{label, counter})
+		if err != nil {
+			return fmt.Errorf("[%s] write error: %w", label, err)
+		}
+		counter++
+	}
+}
+
+// monotonicReader reads json messages from the stream, expecting a sequence
+// It will error if it sees a gap or anything unexpected.
+func monotonicReader(t *testing.T, stream io.Reader, label string) error {
+	expected := 0
+	dec := json.NewDecoder(stream)
+	for {
+		var m msg
+		// Read one line
+		err := dec.Decode(&m)
+		if err == io.EOF {
+			return io.EOF
+		}
+		if err != nil {
+			return fmt.Errorf("[%s] read error after seeing %d values: %w", label, expected, err)
+		}
+		if m.label == "" {
+			continue
+		}
+		if m.label != label {
+			return fmt.Errorf("[%s] read label mismatch: got %q, want %q", label, m.label, label)
+		}
+		if m.val != expected {
+			return fmt.Errorf("[%s] read value mismatch: got %d, want %d", label, m.val, expected)
+		}
+		expected++
+	}
+}
+
+func makeConnPair() (*pipeConn, *pipeConn) {
+	read1, write1 := io.Pipe()
+	read2, write2 := io.Pipe()
+	conn1 := &pipeConn{reader: read1, writer: write2}
+	conn2 := &pipeConn{reader: read2, writer: write1}
+	return conn1, conn2
+}
+
+func failNow(errCh chan error, message string, err error) {
+	errCh <- fmt.Errorf("%s: %w", message, err)
+}
+
+// TestTimeoutParallel verifies that if one yamux stream times out on write,
+// it automatically closes the entire session. This prevents leaving the
+// remote side in a deadlocked read if the bug is present.
+//
+// Steps:
+//  1. Create yamux client/server over a pair of pipeConn with small
+//     ConnectionWriteTimeout and KeepAliveInterval so we can trigger the
+//     scenario quickly.
+//  2. We start two parallel streams: stream A and stream B, each in separate
+//     goroutines. Each writes an infinite sequence of incrementing JSON
+//     objects to the other side, which also reads them in an infinite loop
+//     checking correctness.
+//  3. We artificially pause A's writer. Then, we BLOCK the pipeConn so that
+//     B's writes will fill up the Session's channel and time out on the
+//     server side with ErrConnectionWriteTimeout. We do *not* close the
+//     session or stream after that, as that reflects common usage of this
+//     library.
+//  4. We unblock the pipe. Meanwhile, B's reader on the client side is left
+//     waiting for data that never arrives (under the bug). The read never
+//     completes.
+//
+// If the bug is fixed (i.e., the code forcibly closes the entire session when
+// Stream.Write times out), then the client read on B would quickly see an EOF/reset
+// instead of hanging forever.
+func TestTimeoutParallel(t *testing.T) {
+	// used for waiting at the end of the test
+	var wg sync.WaitGroup
+	// used for failing now
+	errCh := make(chan error, 1)
+	t.Parallel()
+
+	// 1. Setup net.Pipe
+	serverSide, clientSide := makeConnPair()
+
+	// 2. Make yamux config with small timeouts so we see a quick repro
+	serverConf := DefaultConfig()
+	serverConf.ConnectionWriteTimeout = testConnWriteTimeout
+	serverConf.KeepAliveInterval = testKeepAlive
+
+	clientConf := DefaultConfig()
+	clientConf.ConnectionWriteTimeout = testConnWriteTimeout
+	clientConf.KeepAliveInterval = testKeepAlive
+
+	// 3. Start server session
+	serverSes, err := Server(serverSide, serverConf)
+	if err != nil {
+		t.Fatalf("yamux server: %v", err)
+	}
+	defer serverSes.Close()
+
+	// 4. Start client session
+	clientSes, err := Client(clientSide, clientConf)
+	if err != nil {
+		t.Fatalf("yamux client: %v", err)
+	}
+	defer clientSes.Close()
+
+	// 5. Create two streams from server->client: A and B
+	// 6. On the client side, Accept them
+	// 7. We'll run two sets of goroutines: each side has a writer and a reader.
+
+	// A signals for "pause/resume"
+	pauseA := make(chan struct{}, 1)
+
+	// Start A writer (server side)
+	go func() {
+		wg.Add(1)
+		defer wg.Done()
+		streamAserver, err := serverSes.Open()
+		if err != nil {
+			t.Fatalf("server open stream A: %v", err)
+		}
+		err = monotonicWriter(t, streamAserver, "A", pauseA)
+		if err != nil && !errors.Is(err, ErrStreamClosed) {
+			failNow(errCh, "[A-writer-server] received unexpected error", err)
+		}
+	}()
+
+	// Start A reader (client side)
+	go func() {
+		wg.Add(1)
+		defer wg.Done()
+		streamAclient, _ := clientSes.Accept()
+		defer streamAclient.Close()
+		err = monotonicReader(t, streamAclient, "A")
+		if err != nil && !errors.Is(err, io.EOF) {
+			failNow(errCh, "[A-reader-client] received unexpected error", err)
+		}
+	}()
+
+	// Start B writer (server side)
+	go func() {
+		wg.Add(1)
+		defer wg.Done()
+		streamBserver, _ := serverSes.Open()
+		err = monotonicWriter(t, streamBserver, "B", nil)
+		if err != nil && !errors.Is(err, ErrConnectionWriteTimeout) {
+			failNow(errCh, "[B-writer-server] received unexpected error", err)
+		}
+	}()
+
+	// Start B reader (client side)
+	go func() {
+		wg.Add(1)
+		defer wg.Done()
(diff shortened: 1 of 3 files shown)


schmichael (commented):
I think I see what's going on here: the writing side of unidirectional Stream (`streamBserver` in this case) **must** close its connection for the reading side to be notified it has gone away. Your change makes it so _Sessions_ close themselves on any send errors and appears to work as designed.

In theory yamux supports retrying Writes on Streams which receive a `ErrConnectionWriteTimeout` error. However in practice I highly doubt anyone does this because the Write that timed out is still in the sendCh buffer and could succeed later. What is a Stream to do? Wait for a Response just in case the send succeeds? Try the Write again if the application layer protocol is idempotent? I highly doubt any consumers of yamux do anything but `Stream.Close()` on _any_ error from a Stream operation, and I think that's probably for the best.

## This PR

I'm still hesitant to accept this PR because I don't think it's fixing a bug as much as being defensive against coding errors (a missing `Stream.Close()` in this case). I could not find an example of this coding error in Nomad, but I intend to keep looking. We do take these things very seriously!

Speaking of Nomad, another reason I'm hesitant to close the whole Session whenever a single send times out is that Nomad doesn't always know what kind of `net.Conn` its using. It might be a TCP connection, a TLS session, or a yamux Stream. If a Write to any of these connections fail, Nomad will want to explicitly close the connection and create a new one. For example [Go's `crypto/tls` says](https://pkg.go.dev/crypto/tls#Conn.SetDeadline):

> After a Write has timed out, the TLS state is corrupt and all future writes will return the same error.

So clearly the connection is useless, but do you need to explicitly close it? The examples do, and I have to imagine always explicitly closing a connection that will not be reused is the most robust choice.

So if users should be explicitly closing non-yamux.Stream net.Conns on error, I'd prefer they stick to that idiom with yamux.Streams as well.

And finally I think there are some risks to throwing away a whole Session on a single Stream's error. In your test Stream A is fine because it doesn't Write during the partition. But with your code change, Stream A, and any other Streams on a Session that experiences any kind of error, get thrown away. This could cause a lot of unnecessary disruption. For example in Nomad blocking queries may sit in a Read on an idle connection for minutes. The heartbeat mechanism is what ensures those connections don't miss a disconnect, and as far as I can tell the heartbeat mechanism isn't in question.

## Test Variants

To arrive at these conclusions I spent quite a lot of time altering your very useful test case. You can see 2 variants in this gist: https://gist.github.com/schmichael/85bf905c95e0c7b03487b27832e8351c

The first tries to be a minimal variant with 2 changes:
1. It closes streams on errors.
2. It periodically prints received messages to demonstrate Stream A unpauses/unblocks as expected.

The second variant is much more drastic and tries to represent a more idiomatic approach to using yamux. It was extremely useful to play around with, but I will admit the end state has diverged sufficiently from the original that I may not be exercising the same cases.

## Conclusion

Both test variants lead me to believe that *in the absence of coding errors* yamux and tools that use it like Nomad do not risk DoS in the face of transient network partitions. Either heartbeat failures or explicit Close calls on timed out Streams will eventually free both sides of a connection.

Next week I intend to both audit Nomad's use of yamux Streams to ensure we are explicitly closing Streams on errors.

In addition the Nomad team intends to do more testing of full Client and Server agents that communicate through a proxy of our control so we can simulate various flaky conditions.

### THANK YOU!

We take security and reliability of yamux and Nomad seriously and greatly appreciate your contributions! For potential security vulnerabilities we prefer to hear about it directly first like we lay out here: https://www.hashicorp.com/trust/security/vulnerability-management

That triggers a much more robust internal process than "schmichael promises to check out that suspicious issue someday (if he doesn't get distracted first)." :grimacing: 

I also know you've been battling https://github.com/hashicorp/nomad/issues/23305 for a long time, and hopefully this latest attempt will help us find the root cause!

Let me know what you think, and expect to hear back from us on this PR or the original issue soon.

lattwood:
Will adjust the PR based on your feedback when I'm back from PTO.

I realized the security implications after filing the issue, and this was
the response I was given after asking about the status, a week after I got
an initial response, which seemed to indicate they either didn't see the
issue, or (mistakenly) thought that mTLS mitigated it-

```
Thanks again for reaching out and all the details in your PR and GitHub
issue.

This issue was also identified in the past but due to mitigating control
(use of mTLS in clusters following HashiCorp guidance / good practices &
the Nomad security model
<https://developer.hashicorp.com/nomad/docs/concepts/security>), this has
been handled as a longer-lived security / defense-in-depth improvement
rather than a vulnerability that we had in our roadmap.

I'll revive the topic with the Nomad team and have your PR reviewed and
hopefully merged in a future release.

Thanks a lot for the contribution, it's greatly appreciated.
```

So in this case I don't think the more in depth review was triggered via
that mechanism. In fact, it was the presence of mTLS that made me realize
the missing write was never making it onto the wire and I started looking
into the issue from the write instead of read perspective.

lattwood:
Go stdlib issue opened for `net/rpc` error issues- https://github.com/golang/go/issues/71559

tgross:
@lattwood just wanted to give you a heads up that I'm working on building out a reproduction of the Nomad issue in hopes of getting some empirical insights into the underlying problem here. I'll report back on that soon.

lattwood:
This is how I reconcile the broken nodes with pending allocations. This doesn't capture nodes that have a hung read and haven't had an allocation change state.
```
parallel -j16 'ssh -F /dev/null -i ~/.ssh/REDACTED -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@{} '"'killall -9 nomad'" ::: `nomad job status -json | jq -r '.[].Allocations[] | select(.ClientStatus == "pending") | .NodeName' | sort -u`
```

This is how I get the node_id's for everything with a hung read in yamux. I get goroutine stacks via the pprof endpoint, then use overlapping grep commands to find any hosts that have been blocked in yamux for at least 10 minutes.
```
nomad node status -json | jq -r '.[] | select(.Status=="ready") | .ID' | parallel -j16 "curl -s -k --cacert $NOMAD_CACERT --cert $NOMAD_CLIENT_CERT --key $NOMAD_CLIENT_KEY --header X-Nomad-Token:\ $NOMAD_TOKEN https://REDACTED:4646/v1/agent/pprof/goroutine\?debug=2\&node_id={} | grep -B2 yamux | grep -A2 -E ' [0-9]{2,} minutes' | grep -B1 Read | grep minutes > {}.out"
```

From the node_id's, I look them up w/ `nomad node status`, ssh in and kill nomad again.

tgross:
Ok, I'm at the end of my day here but I want to brain-dump what I've got so far...

I don't have a reproduction yet but I have some insights as to why this has been challenging to reproduce. My test rig is a server and a client each in their own network namespace. The structure of the test is to induce traffic from one to the other, and then intervene with either iptables or [`tc`](https://www.man7.org/linux/man-pages/man8/tc.8.html) to simulate network splits/congestion. 

(I also discovered another strange bug in the Nomad client where with iptables split I was able to induce the client to get stuck on a much shorter "wait" for it's blocking RPC than the 5min it should be waiting. We set a shorter wait when retrying and must not be resetting that properly. I'll get an issue open in the Nomad repo for that in the next day or so.)

Inducing the `ErrConnectionWriteTimeout` instead of some other error has proven surprisingly hard, and I think that's because of buffering at the TCP level. The congestion is on the network, but the write timeout we're worried about is in yamux. So if the underlying TCP connection buffer is big enough for yamux to write the header+body to it (via the msgpack codec), then we return without error to yamux's write. Nothing else (in Nomad) will write on that stream till we get a response. But the body can just be sitting in the TCP buffers at that point.

To give myself the best odds of detecting this bug we care about here I'm:
* Setting Nomad's yamux keepalive window to 120s.
* Setting Nomad's yamux write timeout to 1s.
* Setting the TCP connection's write buffer to 1 byte.
* Adding logging to yamux so that it reports sending keepalive, so I can time when to interrupt traffic

After a dozen variations I've managed to trigger the `ErrConnectionWriteTimeout` at [`session.go#L441`](https://github.com/hashicorp/yamux/blob/v0.1.2/session.go#L441) but not the one earlier in the `waitForSendErr`. Unfortunately that test did _not_ trigger the condition where `(*Stream).Read` hangs longer than the network interruption. Once connectivity is restored, the read unblocks.

However, the RPC handler is _not_ getting an error after the `ErrConnectionWriteTimeout`; it's hanging until either the keepalive or the netsplit heals! Which is certainly a bug even if it doesn't quite match the behavior described in #142. I think your observation about `net/rpc` is likely correct, but that would be a bug in the msgpack library and we should fix it there by closing on error as was done in gob (ref https://github.com/golang/go/issues/7689). But before we go running off to do that, I'd really like to have an exact repro of the behavior you're seeing so we can be confident we've put this bug to rest. I'll take another run at it tomorrow and will report back after that and/or open a PR in the msgpack library. 

(Please hold off on opening an issue/PR there, as another team who isn't in the loop on this bug and is in a far-flung TZ will get notifications and that'll cause them a bit of scramble. We're making progress.)

lattwood:
Awesome. If you can find yourself on the community "halihax" tomorrow, I
can also grab arbitrary goroutine stacks, poke around hung agents with
delve, etc.

lattwood:
> After a dozen variations I've managed to trigger the ErrConnectionWriteTimeout at [session.go#L441](https://github.com/hashicorp/yamux/blob/v0.1.2/session.go#L441) but not the one earlier in the waitForSendErr. Unfortunately that test did not trigger the condition where (*Stream).Read hangs longer than the network interruption. Once connectivity is restored, the read unblocks.

Yeah, this is the insidious part. You need to get Write Timeouts on the connection, and then *never* write to it again, to replicate the behaviour of the message pack codec.

https://github.com/hashicorp/yamux/blob/9e980a5a29bca84539fa6271239da99f8962fa38/session.go#L109

What you need to do (which sounds _insane_) is fill that buffered channel to reproduce the hanging read, and have it kept full for the entire Write timeout. It looks like you managed to get the OS buffers full but not yamux's, so *almost* there!

I'm going to open up some PRs to fix *_their_* codec behaviour 🤦🏻‍♂️ , and I'll create a branch and link it here for the msgpack stuff if you don't get to that before I do. I won't actually create the PR until later due to the aforementioned TZ stuff.

tgross:
> I'm going to open up some PRs to fix their codec behaviour 🤦🏻‍♂️ , and I'll create a branch and link it here for the msgpack stuff if you don't get to that before I do. I won't actually create the PR until later due to the aforementioned TZ stuff.

"Their" who? The msgpack codec repo? That's just as much "us" as this one in terms of implementation expertise. But there's another "IP compliance" team that gets pinged on these repos, especially when there's an industry-wide scramble created by the spurious vuln DB filing. Please just hold on, @lattwood, and let us work here.

lattwood:
@tgross ack.

if you're curious-

https://chatgpt.com/share/67a38308-f618-800e-b1ea-6df285d2f2cf

lattwood:
One other realization- it's been stated numerous times that using mTLS with Nomad prevents direct exploitation of this issue.

But starting with 1.10, official guidance is to turn off mTLS when using Workload Identity with Vault.

tgross:
> But starting with 1.10, official guidance is to turn off mTLS when using Workload Identity with Vault.

_For HTTP_, not RPC.

tgross:
I've got a follow-up of where we're at on this @lattwood but give me a little bit to write it up and vet it with @schmichael. Will post back here shortly.

tgross:
I've been working to build out a reproduction of the behavior you've described and I'm still unsuccessful after 2 days.

With a hacked-up API client that starts a stream (via `alloc exec`), pauses with it open long enough for me to create network congestion, and then sends an endless stream of base64 garbage, I can regularly trigger the `ErrConnectionWriteTimeout` as described. But this isn't resulting in the client getting stuck no matter how I tune buffers and timeouts. In all cases, the writer closes either the stream or session (depending on details). Where the reader side doesn't close the stream it's returned to the pool where the window gets resized.

Added logging in `yamux`, `go-msgpack`, and `net-rpc-msgpackrpc` suggests that I'm still not hitting the code path you've identified where `net/rpc` swallows the error. Because of code generation in release builds it gets a little hairy exactly which code path we take to `WriteResponse`, as both libraries implement it.

That being said, the `gob` fix in https://github.com/golang/go/commit/cd5b785efe3cf67a8e065f8ff2ee8a39badbec40 also seems reasonably safe. This also lines-up with libp2p's yamux fork not auto-closing-on-errors: they don't use our msgpack RPC, so they don't have this problem! This problem is only likely to be observed in rpc implementations like this where the actual `Stream.Write` is buried in a blackbox like `net/rpc`. (Also note that libp2p's [may not be protocol-compatible](https://github.com/libp2p/go-yamux/issues/103#issuecomment-1511915380), so just dropping it in may be a dead end.)

So here's what I'd like to propose:
* Go ahead and open the `gob` fix as PRs on the `go-msgpack` and `net-rpc-msgpackrpc` libraries. Reference https://github.com/hashicorp/nomad/issues/23305 and tag @schmichael and I in the description. We'll pop in and let other HashiCorp folks who might jump in know that this is work-in-progress.
* Build yourself a binary with that fix and deploy it to some chunk of your cluster.
* If the problem goes away, great! We'll merge the PRs and ship them.
* If not, lets head back to [#23305](https://github.com/hashicorp/nomad/issues/23305) and @schmichael and I will help debug from there.
  * Bring data! Ideally we'll get logs and goroutine stacks from both ends of the problem.
  * If it becomes possible for you to provide a minimal reproduction, that'd be even better.
  * Let's try to keep the discussion centralized in [#23305](https://github.com/hashicorp/nomad/issues/23305) until we have that smoking gun. My gut tells me that the symptoms you're having may extend beyond networking/RPC issues, so it'll help if we can dissect things together in one place.

PR #147, opened 2025-03-03
Coverage test and Linting

KaushikiAnand:
- I have added unit test coverage and linting to the test.yml workflow. The .go-version file also added. 

- The unit test coverage change was done to help understand the defects early in lifecycle and come up with a good solution for it. The linting was done to ensure that the codebase is consistent and maintainable and helps in improving the code quality and reducing errors. The .go-version file was added to find the security vulnerabilities in a repo.


.go-version
```diff
@@ -0,0 +1 @@
+1.20
```
tgross, line 1: This should probably be the oldest supported version of Go, rather than an unsupported version. We should consider the same for the setup-go step in the workflows as well.

KaushikiAnand: can i get a bit more understanding on which go-version are you expected to be present ?

tgross: See https://go.dev/doc/devel/release#policy

KaushikiAnand: suggested change made

tgross (approved):
LGTM

mohanmanikanta2299 (approved):

PR #150, opened 2025-06-11
IND-3879 enabling dependabot, changelog.md added

KaushikiAnand:
- As per the memo: https://go.hashi.co/memo/eng-004,
Enabling the dependabot for identifying and updating the security and version updates in the repositories rather handling it manually.
- Adding the CHANGELOG.md file to maintain standardisation across OSS core libraries


abhijeetviswa (approved):

.github/dependabot.yaml
```diff
@@ -0,0 +1,34 @@
...
+      prefix: "[chore] : "
+    groups:
+      go:
+        patterns:
+          - "*"
+        applies-to: "version-updates"
+      go-security:
+        patterns:
+          - "*"
+        applies-to: "security-updates"
```
tgross, line 34: Missing EOL. It doesn't matter for this file but you should fix your editor so that it always ensures you've got one on save. Otherwise you end up with this:

```
wc -l .github/dependabot.yaml
  33
```

tgross (approved):

abhijeetviswa (approved):
There is a failing test due to a network timeout. Please investigate and raise a fix for it in a separate PR.

Issue #116, opened 2022-12-23
What is Disconnect Flow


kudla:
Please, what is intended graceful disconnection flow?
* Should a client it-self initiate a disconnection or should it be controlled on higher level and both server and client should be closed?
* Are client and server enough to be closed or subsequent streams should be closed first
* Same who should be responsible for the streams. to be  closed  (client/server/controlled both)

Trying to close just a client side leads to `EOF`s and `keepalive timeouts`.  So is there design supposed scenario or I've to manage it on my app explicitly?

Thx

andig:
Similar issue here. Currently, I'm disconnecting the client using `client.Close()`, then close the websocket. This in turn leads to

    [ERR] yamux: Failed to read header: failed to get reader: failed to read frame header: EOF

in the server log which is confusing.

PR #167, opened 2026-04-13
Fix yamux: Failed to read header server error

andig:
## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.

The `recvLoop` code checks websocket errors with `err != io.EOF`. This is ancient Go and not best practice. 
When using `coder/websocket` as underlying library as in our case, closed socket streams wrap errors (see https://github.com/coder/websocket/blob/d099e1621e8d0f080c9f3b87c5a8587e0b722723/frame.go#L52-L53):

```go
// Wrap wraps err with fmt.Errorf if err is non nil.
// Intended for use with defer and a named error return.
// Inspired by https://github.com/golang/go/issues/32676.
func Wrap(err *error, f string, v ...any) {
	if *err != nil {
		*err = fmt.Errorf(f+": %w", append(v, *err)...)
	}
}
```

To catch this error and guard against logging it the error comparison needs to use `errors.Is()` which is also modern Go. 

This takes care of https://github.com/hashicorp/yamux/issues/116#issuecomment-4224054753.

### DIFF
--- a/session.go
+++ b/session.go
@@ -555,7 +555,7 @@ func (s *Session) recvLoop() error {
 		// Read the header
 		if _, err := io.ReadFull(s.bufRead, hdr); err != nil {
 			if !errors.Is(err, io.EOF) && !strings.Contains(err.Error(), "closed") && !strings.Contains(err.Error(), "reset by peer") {
-				s.logger.Printf("[ERR] yamux: Failed to read header: %v (%T)", err, err)
+				s.logger.Printf("[ERR] yamux: Failed to read header: %v", err)
 			}
 			return err
 		}
--- a/session.go
+++ b/session.go
@@ -7,6 +7,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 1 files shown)


tgross:
Hey @andig there's no websockets involved here anywhere at all whatsoever in this library, so your explanation feels a little... random. #116 hasn't even been accepted as a bug; it's just an open question which we need to get to. What does this actually fix?

andig:
Thanks for asking, sorry I wasn't clear (and it's not a full fix yet as I realize). 

We're seeing this in the tunnel server's log:

    [ERR] yamux: Failed to read header: failed to get reader: failed to read frame header: EOF

when the client closes the session via `client.Close()`. We're using [coder/websocket](https://github.com/coder/websocket) as underlying transport:

```go
conn, _, err := websocket.Dial(ctx, t.tunnelURL, &websocket.DialOptions{})
if err != nil {
    return false, fmt.Errorf("websocket dial: %w", err)
}

netConn := websocket.NetConn(ctx, conn, websocket.MessageBinary)

session, err := yamux.Client(netConn, yamux.DefaultConfig())
```

Goal is to get rid of the server log message that is expected. The issue is the duck-typing on the error plus the text check for the error message. It seems we'd need to get rid of both.

Does that make sense?

andig:
If it does make sense: what kind of errors would you want to log here- anything other than an EOF? Then we could also remove the string checking. Btw., I realize that:

```go
// EOF is the error returned by Read when no more input is available.
// (Read must return EOF itself, not an error wrapping EOF,
// because callers will test for EOF using ==.)
// Functions should return EOF only to signal a graceful end of input.
// If the EOF occurs unexpectedly in a structured data stream,
// the appropriate error is either [ErrUnexpectedEOF] or some other error
// giving more detail.
var EOF = errors.New("EOF")
```

...but that doesn't happen in the wild as far as I can see.

andig:
Confirmed PR working and opened upstream issue: https://github.com/coder/websocket/issues/561

